PR #111, opened 2023-12-11
[pacman] 6.0.2-10: enable lto

YukariChiba:


YukariChiba:
RISC-V support for gold is not ready. Disable LTO for RISC-V only.

Ref: https://discussion.fedoraproject.org/t/fedora-riscv-should-we-disable-lto-gcc-already-disabled-for-clang/96582

PR #121, opened 2024-01-22
[busybox] 1.36.1-14: add and fix mdev helpers

YukariChiba:


YukariChiba:
@YukariChiba mouse hotplug works on laptop (livecd)

PR #440, opened 2024-06-13
[pacman] 6.0.2-27: add --no-as-needed to riscv64 ldflags

YukariChiba:


ziyao233:
we now have better choice to fix riscv64 port

PR #507, opened 2024-07-02
[pcre] 8.45-3: disable JIT on riscv64

ziyao233:
JIT is not supported on riscv64


ziyao233:
should not affect other architecture, check buildlog to make sure JIT is enabled on x86_64 and aarch64 before merging,

YukariChiba (approved):

PR #570, opened 2024-07-09
[libseccomp] 2.5.5-2: disable test 52 to fix qemu-user and riscv64

YukariChiba:


ziyao233 (commented):
it is better to comment the reason of change, since it is a workaround but not a real fix

PR #1204, opened 2024-09-06
[winetricks] 20240105-1: init package

YukariChiba:


PKGBUILD
```diff
@@ -0,0 +1,21 @@
+# Maintainer: Yukari Chiba <i@0x7f.cc>
+
+pkgname=winetricks
+pkgver=20240105
+pkgrel=1
+pkgdesc='Script to install various redistributable runtime libraries in Wine.'
+url='https://wiki.winehq.org/winetricks'
+license=('LGPL')
```
ziyao233, line 8: Use spdx identifier here (LGPL-2.0-or-later or something similar)

ziyao233 (changes requested):

PR #1567, opened 2024-10-02
[android-tools] 35.0.2-1: new upstream version

ziyao233:


ziyao233:
android-tools 35.0.2 comes with usage of unreleased libusb APIs, let's delay the update.

Reference: https://github.com/nmeum/android-tools/issues/154

PR #1676, opened 2024-10-10
[fcitx5-anthy] 5.1.4-2: fix arch

ziyao233:


PKGBUILD
```diff
@@ -5,7 +5,7 @@ pkgver=5.1.4
 pkgrel=1
```
YukariChiba: bump pkgrel

YukariChiba (changes requested):

PR #1830, opened 2024-10-20
[python] 3.12.7-3: fix makedepends, various cleanup

ziyao233:
Add linux-headers to makedepends to fix missing os.*xattr().


ziyao233:
composefs should be rebuilt after merging

PR #2697, opened 2024-12-26
[perl] 5.40.0-3: fix path for perl

YukariChiba:


PKGBUILD
```diff
@@ -22,13 +26,18 @@ build()
...
+    -Darchlib=/usr/lib/perl5/$_baseversion/core_perl \
+    -Dsitelib=/usr/share/perl5/site_perl \
+    -Dsitearch=/usr/lib/perl5/$_baseversion/site_perl \
+    -Dvendorlib=/usr/share/perl5/vendor_perl \
+    -Dvendorarch=/usr/lib/perl5/$_baseversion/vendor_perl \
     -Dman1dir=/usr/share/man/man1 \
     -Dman3dir=/usr/share/man/man3 \
+    -Dscriptdir=/usr/bin/core_perl \
+    -Dsitescript=/usr/bin/site_perl \
+    -Dvendorscript=/usr/bin/vendor_perl \
```
ziyao233: I'll prefer to keep the execs in the default path and drop perlbin.sh, which will cause perl to install its binaries (including perl interpreter/cpan) to `/usr/bin/core_perl`. This'll break chroot-based environment (they don't respect /etc/profile.d) and I don't see its value.

ziyao233 (approved):

PR #2792, opened 2024-12-28
[npm] 11.3.0-1: init package

YukariChiba:


PKGBUILD
```diff
@@ -0,0 +1,75 @@
+# Maintainer: Yukari Chiba <i@0x7f.cc>
+
+pkgname=npm
+pkgver=11.3.0
+pkgrel=1
+pkgdesc='JavaScript package manager'
+arch=(any)
+url=https://www.npmjs.com
```
ziyao233: Quote the string

PKGBUILD
```diff
@@ -0,0 +1,75 @@
...
+  node-gyp
+  nodejs
+  nodejs-nopt
+  semver
+)
+# symlinks for ln -r
+makedepends=(git symlinks)
+optdepends=("git: for dependencies using Git URL's")
+source=("npm-cli::git+https://github.com/npm/cli.git#tag=v$pkgver")
+sha256sums=('SKIP')
```
ziyao233: It's okay to specify a sha256sum, we have a dedicated tag to fetch here.

PKGBUILD
```diff
@@ -0,0 +1,75 @@
...
+
+build() {
+  cd npm-cli
+  node scripts/resetdeps.js
+}
+
+check() {
+  cd npm-cli
+  # Windows shims test failure
+  mv test/bin/windows-shims.js{,.bak}
```
ziyao233: Could we simply remove the test? It's only a test script.

PKGBUILD
```diff
@@ -0,0 +1,75 @@
...
+  for page in man5/folders.5 man5/install.5 man7/*.7; do
+    sec=${page##*.}
+    name=$(basename "$page" ."$sec")
+    title=${name@U}
+
+    sed -Ei "s/^\.TH \"$title\"/.TH \"NPM-$title\"/" "$page"
+    mv "$page" "${page/\///npm-}"
+  done
+
+  gzip man?/*
```
ziyao233, line 64: IMHO this should be unnecessary as pacman is able to compress man-pages, right?

ziyao233 (changes requested):

PKGBUILD
YukariChiba: actually they're located at `usr/lib/node_modules/npm/man/man*/*` and symlinked to `usr/share/man/man*/*`

ziyao233 (approved):

PR #2915, opened 2025-01-12
[libgpiod] 2.2-1: new package

Ratizux:
Created package for libgpiod, splitted python binding as `python-gpiod`

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,42 @@
+# Maintainer: Weiji Wang <nebclllo0444@gmail.com>
+
+pkgname=('libgpiod' 'python-gpiod')
+pkgbase='libgpiod'
+pkgver=2.2
+pkgrel=1
+pkgdesc='C library and tools for interacting with the linux GPIO character device'
+url='https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git'
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+license=('LGPL-2.1-or-later')
+depends=('musl')
+makedepends=('autoconf-archive' 'help2man' 'python-setuptools' 'python-build' 'linux-headers')
+source=("https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/snapshot/libgpiod-$pkgver.tar.gz"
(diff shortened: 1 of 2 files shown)


PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+pkgbase='libgpiod'
+pkgver=2.2
+pkgrel=1
+pkgdesc='C library and tools for interacting with the linux GPIO character device'
+url='https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git'
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+license=('LGPL-2.1-or-later')
+depends=('musl')
+makedepends=('autoconf-archive' 'help2man' 'python-setuptools' 'python-build' 'linux-headers')
+source=("https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/snapshot/libgpiod-$pkgver.tar.gz" '0001-disable-python-build-isolation.patch')
```
ziyao233: nit: split into 2 lines

PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+	_patch_ "$pkgname-$pkgver"
+}
+
+build() {
+	cd "$pkgname-$pkgver"
+	./autogen.sh \
+		--prefix=/usr \
+		--enable-tools=yes \
+		--enable-bindings-cxx \
+		--enable-bindings-python
```
ziyao233, line 28: make sure building is done in `build()` stage, instead of relying on the implicit dependency tracking of Makefile.

PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+}
+
+package_libgpiod() {
+	depends=('musl-static')
```
ziyao233: Does it really depend on a static package? Why?

Ratizux: When using libgpiod as C/C++ library, the header includes math.h which raises an undefined reference without `musl-static` installed

ziyao233: musl-static doesn't ship any headers, I guess the error is thrown during linking?

musl-static is part of base-devel, so it should never be there.

ziyao233 (approved):

PR #2982, opened 2025-01-15
[ethtool] 6.11-1: init package

JulianDroske:
currently only x86_64 is tested

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,42 @@
+# Maintainer: Julian <juliandroske@outlook.com>
+
+pkgname=ethtool
+pkgver=6.11
+pkgrel=1
+pkgdesc="Utility for controlling network drivers and hardware"
+url="https://cdn.kernel.org/pub/software/network/ethtool"
+arch=(x86_64 aarch64 riscv64 loongarch64)
+license=(GPL-2.0-only)
+depends=(
+  musl
+  libmnl
+)
+makedepends=(
+  linux-headers
+)
+source=(
+  "https://cdn.kernel.org/pub/software/network/$pkgname/$pkgname-$pkgver.tar.gz"
+)
+sha256sums=('a8bee475d991939cad182b7dbcf3a1f139aeed8cb0add73d72084cd71d43282b')
+
+prepare() {
+  _patch_ "$pkgname-$pkgver"
+}
+
+build() {
+  cd "$pkgname-$pkgver"
+  ./configure --prefix=/usr \
+    --mandir=/usr/share/man \
+    --sbindir=/usr/bin
+  make
+}
+
+check() {
+  cd "$pkgname-$pkgver"
+  make check
+}
+
+package() {
+  cd "$pkgname-$pkgver"
+  make DESTDIR="$pkgdir" install
+}


PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: Julian <juliandroske@users.noreply.github.com>
+
+pkgname=ethtool
+pkgver=6.11
+pkgrel=1
+pkgdesc="Utility for controlling network drivers and hardware"
+url="https://cdn.kernel.org/pub/software/network/ethtool"
+arch=(x86_64)
+license=(GPL-2.0-only)
+options=(!lto)
```
ziyao233: Why disable lto?

PKGBUILD
```diff
@@ -0,0 +1,41 @@
+# Maintainer: Julian <juliandroske@users.noreply.github.com>
+
+pkgname=ethtool
+pkgver=6.11
+pkgrel=1
+pkgdesc="Utility for controlling network drivers and hardware"
+url="https://cdn.kernel.org/pub/software/network/ethtool"
+arch=(x86_64)
```
ziyao233: Please add all architectures (`x86_64` `aarch64` `riscv64` `loongarch64`), or it'll never be tested.

PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+license=(GPL-2.0-only)
+options=(!lto)
+depends=(
+  libmnl
```
ziyao233, line 12: Add `musl` to `depends`, if it isn't linked statically (it shouldn't be)

PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+)
+sha256sums=('a8bee475d991939cad182b7dbcf3a1f139aeed8cb0add73d72084cd71d43282b')
+
+prepare() {
+  _patch_ ethtool-$pkgver
+}
+
+build() {
+  cd "$srcdir/$pkgname-$pkgver"
+  ./configure --prefix=/usr --mandir=/usr/share/man --sbindir=/usr/bin
```
ziyao233: nit: format it to ease reader's eyes:

```suggestion
  ./configure --prefix=/usr \
      --mandir=/usr/share/man  \
      --sbindir=/usr/bin
```

PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+  "https://cdn.kernel.org/pub/software/network/$pkgname/$pkgname-$pkgver.tar.gz"
+)
+sha256sums=('a8bee475d991939cad182b7dbcf3a1f139aeed8cb0add73d72084cd71d43282b')
+
+prepare() {
+  _patch_ ethtool-$pkgver
+}
+
+build() {
+  cd "$srcdir/$pkgname-$pkgver"
```
ziyao233: drop `$srcdir`. also applies for `check()` and `install()`

PKGBUILD
```diff
@@ -0,0 +1,42 @@
+# Maintainer: Julian <juliandroske@users.noreply.github.com>
```
ziyao233: It's better to leave a valid e-mail.

ziyao233 (changes requested):

PKGBUILD
JulianDroske: I thought it may be needed on other platforms... ref: busybox, but actually it can run without setting this.
Should i remove it?

ziyao233: busybox makes heavy use of compiler tricks and LTO leads to miscompilation. So is disabled. If this isn't `ethtool`'s case, please enable it (just remove the line)

ziyao233 (approved):

ziyao233:
Thanks for your contribution!

PKGBUILD
JulianDroske: Fixed.

PR #2983, opened 2025-01-15
[busybox] 1.37.0-8: add support for networkmanager & fix mdev helper mac addr bug

JulianDroske:
**PR hanged up due to**

* NM won't use a nic without receiving a event, even when mdev has already notified it.
  * Investigating

---

Summary:

* Fixes "infinite increasing interface names" bug, will get original mac address if either ethtool or iproute2 is installed.
* Reverts 9806fe4
* Will notify NetworkManager if possible (i.e. after libudev-zero is installed)


ziyao233:
Excellent work! So in conclusion, for solving issue of unstable interface names, we need to

1. Remove the downstream patch[1] to make networkmanager wait for netcards being configured
2. Apply this PR, to
  2.1 avoid `mdev-helper-settle-nics` misreading a random-generated MAC address
  2.2 generate ready events for netcards, which networkmanager listens to

Here are my questions,

1. Is 2.1 really necessary? If networkmanager waits for `mdev-helper-settle-nics` finish and only sets a random MAC address after it, the helper shouldn't see the spoofed address.
2. What's the exact type of event generated in 2.2? I've seen `libudev-helper` send a NETLINK message, but not sure who's interesting in it: is it a networkmanager-specific event, a kernel-defined event that libudev-zero cares about, or something else?

[1]: https://github.com/eweOS/packages/blob/9eceb0d729f6677c3c1d71a8c30e0f610c5310b2/force-platform-init.patch

mdev-helper-settle-nics
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
```
ziyao233, line 1: No action required. Actually we should fix this: this script makes use of bash features (`local` for example).
We could create a split repository to maintain these helpers later.

mdev-helper-settle-nics
```diff
@@ -28,7 +29,9 @@
 # First it will run nameif to rename interfaces if configured in /etc/mactab
 # Then it will dump known interfaces to new mactab file.
 # Next it will parse old /etc/mactab and copy all interfaces which wasnt added in step two.
-# On the end, it will replace /etc/mactab with the new one.
+# And it will replace /etc/mactab with the new one, controlled by '--write-mactab'.
```
ziyao233, line 32:
```suggestion
# And it will replace /etc/mactab with the new one if '--write-mactab' is supplied.
```

mdev-helper-settle-nics
```diff
@@ -28,7 +29,9 @@
...
+# Finally it will dispatch a udev event for current nic if executed by mdev, controlled by
+# '--notify-udev'.
```
ziyao233, line 34:
```suggestion
# Finally it will dispatch a udev event for current nic if '--notify-udev' is supplied. This relies on
# a correct environment variable `$SEQNUM`, which mdev sets.
```

mdev-helper-settle-nics
```diff
@@ -128,47 +160,51 @@ for nic in /sys/class/net/*_tmp; do
...
+			esac
+		done < '/etc/mactab'
+	fi
+
+	mv "${tmpfile}" '/etc/mactab'
 fi
 
-mv "${tmpfile}" '/etc/mactab'
+# It's time to dispatch a udev event
+# This only works when in mdev.conf as which sets a necessary envar SEQNUM, we may not set a custom one
```
ziyao233, line 207:
```suggestion
# This is only possible when a correct `SEQNUM` is supplied, which is the case when being invoked from mdev.conf
```

mdev-helper-settle-nics
```diff
@@ -128,47 +160,51 @@ for nic in /sys/class/net/*_tmp; do
...
-			'#'*)
+			*_tmp)
 			;;
-			*)
-				if [ -n "${macaddr}" ] && ! in_comma_list "${macaddr}" "${detected_macs}" && ! in_comma_list "${device}" "${detected_nics}"; then
+				eth[0-9]*|wlan[0-9]*|ath[0-9]*|wifi[0-9]*|ra[0-9]*)
+				macaddr="$(get_nic_macaddr ${device} 2>/dev/null)"
+				if [ -n "${macaddr}" ] && ! in_comma_list "${macaddr}" "${inconf_macs}" && ! in_comma_list "${device}" "${inconf_nics}"; then
+					detected_nics="${detected_nics},${device}"
+					detected_macs="${detected_macs},${macaddr}"
```
ziyao233, line 180: I think this is a copy of original code? Could we split a function out and make it easy for reading?

ziyao233 (changes requested):

ziyao233:
For the commit message, please use

```
Reference: https://github.com/NetworkManager/NetworkManager/commit/4f6c91d
Reference: https://github.com/NetworkManager/NetworkManager/commit/1b49f94
```

instead of `refs:`. Or use `[1]` `[2]` as reference and list these URLs at the end of the message, like what I've done in[1]. This reads better.

Keep the first line short (less than 72 characters), I think `[busybox] 1.37.0-8: Correctly dispatch udev event and use permanent MAC` is appropriate.

[1]: https://github.com/eweOS/packages/pull/2983#issuecomment-2595137015

mdev-helper-settle-nics
JulianDroske: > I think this is a copy of original code?

Yes, it's a wrapper of the original impl of "--write-mactab", and I intended to make it be the same level as that of "--notify-udev", so that the structure will be more clear.

> Could we split a function out and make it easy for reading?

Well, according to the original code, the structure is like:
- declare basic functions
- lock file
- parse nics
- (options) write mactab, notify udev

So I was trying to keep it. But it's okay to split them out.

---

And I also think it's time to refactor it, because it's pretty old and didn't share the same idea as the impl of "--notify-udev". i.e. the latter is intended to be executed only by mdev & expects only one device per instance, while the original code can be run by user & processes all nics.

ziyao233: > > I think this is a copy of original code?
> 
> Yes, it's a wrapper of the original impl of "--write-mactab", and I intended to make it be the same level as that of "--notify-udev", so that the structure will be more clear.
> 
> > Could we split a function out and make it easy for reading?
> 
> Well, according to the original code, the structure is like:
> 
>     * declare basic functions
> 
>     * lock file
> 
>     * parse nics
> 
>     * (options) write mactab, notify udev
> 
> 
> So I was trying to keep it. But it's okay to split them out.

I'm willing to see 3 being split into a function, then you could save some indentation.

> And I also think it's time to refactor it, because it's pretty old and didn't share the same idea as the impl of "--notify-udev". i.e. the latter is intended to be executed only by mdev & expects only one device per instance, while the original code can be run by user & processes all nics.

I would second the idea, but let's make everything works first :)

JulianDroske:
> Is 2.1 really necessary?

Actually.. yes, definitely. The problem was that the script will parse **all** nics on each run. When a second nic is plugged, the script will also **reprocess old nics** which may already be held by networkmanager, thus we may only impl this function, unless we refactor this script.

ziyao233:
> > Is 2.1 really necessary?
> 
> Actually.. yes, definitely. The problem was that the script will parse **all** nics on each run. When a second nic is plugged, the script will also **reprocess old nics** which may already be held by networkmanager, thus we may only impl this function, unless we refactor this script.

Understand, that's reasonable. As for now, let's keep the original structure and workaround this.

ziyao233:
>     2. What's the exact type of event generated in 2.2? I've seen `libudev-helper` send a NETLINK message, but not sure who's interesting in it: is it a networkmanager-specific event, a kernel-defined event that libudev-zero cares about, or something else?

I found the answer: libudev-zero relies on NETLINK for hotplugging support. The device manager should rebroadcast uevents on NETLINK_KOBJECT_UEVENT, then libudev-zero (the library) could know hotplugging is happening.. busybox mdev doesn't support this, so comes `libudev-helper`, which rebroadcasts the event to group 2 (`0x4`, `1 << 2`) on NETLINK_KOBJECT_UEVENT.

ziyao233:
> > ```
> > 2. What's the exact type of event generated in 2.2? I've seen `libudev-helper` send a NETLINK message, but not sure who's interesting in it: is it a networkmanager-specific event, a kernel-defined event that libudev-zero cares about, or something else?
> > ```
> 
> I found the answer: libudev-zero relies on NETLINK for hotplugging support. The device manager should rebroadcast uevents on NETLINK_KOBJECT_UEVENT, then libudev-zero (the library) could know hotplugging is happening.. busybox mdev doesn't support this, so comes `libudev-helper`, which rebroadcasts the event to group 2 (`0x4`, `1 << 2`) on NETLINK_KOBJECT_UEVENT.

Then comes another problem: we don't rebroadcast uevents for most subsystems (now only for `drm` and `input`, and probably `net` if we merge this). At least `usb` should be handled as well.

Is it possible to write a generic rule that make mdev invoke `libudev-helper` on each uevent?

JulianDroske:
> What's the exact type of event generated in 2.2?

alias NM=NetworkManager

Now I'm diving into it. Information below was concluded from my own experience (and source codes).

For NM, only five arguments are used, four of them are checked by libudev-zero [1](https://github.com/illiliti/libudev-zero/blob/bbeb7ad51c1edb7ab3cf63f30a21e9bb383b7994/udev_device.c#L666) and may be used by NM:

- SEQNUM (I remembered it also came from an issue for libudev-zero & nm)
  - It's a number
  - I **haven't** figure out if it should be **unique**, but a random number was tested to work for NM.
- SUBSYSTEM
  - NM only receives =net
- DEVPATH/DEVNAME (either of them should appear)
  - NM uses DEVPATH as nics don't have a /dev/* node so no DEVNAME
- ACTION
  - NM processes at least =add

And one necessary argument is used by NetworkManager [2](https://github.com/NetworkManager/NetworkManager/blob/191aaf0fb2a85708a97ad2a99fd9fc05a6241347/src/core/devices/nm-device-generic.c#L96):

- IFINDEX

NM uses IFINDEX to find nic and manage it.

JulianDroske:
This PR contains fixable bugs as we cannot figure out the behavior of NM & udev for now, so it's closed.

All suggestions under this PR may be usable for further development.

Wait until [libudev-zero supports NM](https://github.com/illiliti/libudev-zero?tab=readme-ov-file#what-doesnt-work).

PR #2985, opened 2025-01-16
[micro] 2.0.14-1: init package

JulianDroske:


PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+depends=(
+  musl
+)
+makedepends=(
+  git
+  go
+)
+optdepends=(
+  'wl-clipboard: for copying/pasting text using Wayland'
+  'xclip: for copying/pasting text using X'
```
ziyao233: drop, eweOS won't ship X11-related packages.

PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+source=(
+  "https://github.com/zyedidia/micro/archive/refs/tags/v$pkgver.tar.gz"
+)
+sha256sums=(
+  '40177579beb3846461036387b649c629395584a4bbe970f61ba7591bd9c0185a'
+)
+
+prepare() {
+  cd "$pkgname-$pkgver"
+}
```
ziyao233: drop

ziyao233 (changes requested):

PKGBUILD
JulianDroske: okay i will keep the wl-clipboard only

PKGBUILD
JulianDroske: Okay~

ziyao233 (approved):

PR #2986, opened 2025-01-16
[linux-firmware] 20250109-1: new upstream version

ziyao233:


ziyao233:
recheck after #2987: new release requires GNU parallel for parallel building.

PR #3004, opened 2025-01-18
[emacs] 29.4-1: init package

JulianDroske:
This is the first version of emacs for eweOS.

Currently only necessary dependencies are included to make porting easier. Actual experience needs to be measured.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,122 @@
+# Maintainer: Julian Droske <juliandroske@outlook.com>
+
+pkgbase=emacs
+pkgname=(
+  emacs
+  emacs-nogui
+)
+pkgver=29.4
+pkgrel=1
+url='https://www.gnu.org/software/emacs/emacs.html'
+pkgdesc='Extensible, customizable, self-documenting free/libre text editor'
+arch=(x86_64 aarch64 riscv64 loongarch64)
+license=('GPL-3.0-or-later AND LGPL-2.1-or-later')
+provides=(emacs)
+conflicts=(emacs)
+depends=(
+  musl
+  acl
+  dbus
+  ncurses
+  libxml2
+  zlib-ng
+)
+makedepends=(
+  make
+  texinfo
+
+  linux-headers
+
+  at-spi2-core
+  libjpeg
+  libpng
+  librsvg
+  dav1d
+  libtiff
+  fontconfig
+  freetype2
+  harfbuzz
+  pango
+  cairo
+  glib
+  gtk3
+  gdk-pixbuf
+  alsa-lib
+)
+source=(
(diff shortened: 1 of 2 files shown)


ziyao233 (approved):

PKGBUILD
```diff
@@ -0,0 +1,128 @@
...
+
+  _postpackage
+}
+
+package_emacs-nogui() {
+  pkgdesc="${_pkgdesc} - without GUI"
+  depends=(
+    "${_runtime_common_deps[@]}"
+  )
+  provides+=(emacs-nox)
```
ziyao233, line 116: drop it? literally the GUI version doesn't use X as well....

PKGBUILD
```diff
@@ -0,0 +1,128 @@
...
+url='https://www.gnu.org/software/emacs/emacs.html'
+pkgdesc="${_pkgdesc}"
+arch=(x86_64 aarch64 riscv64 loongarch64)
+license=('GPL-3.0-or-later AND LGPL-2.1-or-later')
+provides=(emacs)
+conflicts=(emacs)
+depends=(
+  "${_runtime_common_deps[@]}"
+  "${_runtime_gui_deps[@]}"
+)
```
ziyao233: Shouild these two packages be conflict?

ziyao233 (dismissed):

PKGBUILD
JulianDroske: Actually there's no conflict or duplication. The `gui_deps` means all deps that used by GUI but not included by `common_deps`.

I didn't know the real way that used by pacman to find deps, so I made everything declared which made nonsense.

As a result, I will present them directly instead of declaring extra envars, for the next force push.

JulianDroske: Well, since they produce the same executable (command) names, they should conflict with each other. 

Of course we don't know ' if they can coexist ' without a test, so keep it for the moment to avoid problems.

PKGBUILD
JulianDroske: Well, okay~

<details>
<summary>Notes</summary>
<br>

Three weeks ago we talked about it and you said:

> emacs-nogui, provides emacs-nox?

Since all other distros called it `nox`, I added this alias in order to care users transitioned from other distros.

Later we can write a note in the wiki, anyway.

</details>

ziyao233 (approved):

PKGBUILD
ziyao233: > Well, okay~
> Notes
> 
> Three weeks ago we talked about it and you said:
> 
> > emacs-nogui, provides emacs-nox?
> 
> Since all other distros called it `nox`, I added this alias in order to care users transitioned from other distros.
> 
> Later we can write a note in the wiki, anyway.

Yeah, I think a precise wiki page is absolutely better than a confusing name.

ziyao233:
thanks for the work!

PR #3106, opened 2025-02-11
[light] 1.2.2-1: new package

alifnurc:


PKGBUILD
```diff
@@ -0,0 +1,53 @@
+# Maintainer: Daniel Peukert <daniel@peukert.cc>
+# Contributor: Maxim Baz <archlinux at maximbaz dot com>
+# Contributor: Fredrik Haikarainen <fredrik.haikarainen@gmail.com>
+# Contributor: Marcin (CTRL) Wieczorek <marcin@marcin.co>
+pkgname='light'
+pkgver='1.2.2'
+_commit='2a54078cbe3814105ee4f565f451b1b5947fbde0'
+pkgrel='5'
+pkgdesc='A program to control backlights (and other hardware lights)'
+arch=('x86_64' 'aarch64')
```
YukariChiba: use `arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')` in eweOS if packages can be built and used on all architectures.

PKGBUILD
```diff
@@ -0,0 +1,53 @@
+# Maintainer: Daniel Peukert <daniel@peukert.cc>
+# Contributor: Maxim Baz <archlinux at maximbaz dot com>
+# Contributor: Fredrik Haikarainen <fredrik.haikarainen@gmail.com>
+# Contributor: Marcin (CTRL) Wieczorek <marcin@marcin.co>
+pkgname='light'
+pkgver='1.2.2'
+_commit='2a54078cbe3814105ee4f565f451b1b5947fbde0'
+pkgrel='5'
```
YukariChiba: use `pkgrel=1` here, it's the first release in eweOS

light.install
YukariChiba, line 1: Remove it since we don't have the older package.

namespace-udev-rule.diff
YukariChiba, line 1: remove it since we use libudev-zero, udev rules does not work here.
**Remember to test if this package works with eweOS since it installs udev rules**

PKGBUILD
```diff
@@ -0,0 +1,53 @@
...
+
+	patch --forward -p1 < "../fix-global-var.diff"
+	patch --forward -p1 < "../namespace-udev-rule.diff"
+	patch --forward -p1 < "../update-manpage.diff"
+}
+
+build() {
+	cd "$srcdir/$_sourcedirectory/"
+	./autogen.sh
+	./configure --prefix='/usr' --with-udev
```
YukariChiba: remove `--with-udev` since there is not udev daemon in eweOS (we use libudev-zero)

PKGBUILD
```diff
@@ -0,0 +1,53 @@
+# Maintainer: Daniel Peukert <daniel@peukert.cc>
```
YukariChiba: Delete all lines here.
Write your name and email in `Maintainer:` if you want to maintain this package.
Otherwise write yourself in `Contributor: ` and leave `Maintainer:` blank.

YukariChiba (changes requested):

namespace-udev-rule.diff
alifnurc: Thanks, I'll do the fix so it can be compatible with libudev-zero.

PKGBUILD
alifnurc: because its newer version of the package.

PKGBUILD
alifnurc: yeah, but I don't have device for the test, I'll try to add riscv64 and loongarch64.

.SRCINFO
```diff
@@ -0,0 +1,22 @@
+pkgbase = light
```
ziyao233: Please remove `.SRCINFO`, `.editorconfig` and `.gitignore`. We store only the PKGBUILD and necessary patches/package configuration in this repository.

PKGBUILD
ziyao233: > because its newer version of the package.

`pkgrel` is only bumped when a downstream change is made. upstream versions are represented by `pkgver`.

Please read https://wiki.archlinux.org/title/Arch_package_guidelines#Package_versioning

PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+pkgver='1.2.2'
+_commit='2a54078cbe3814105ee4f565f451b1b5947fbde0'
+pkgrel='5'
+pkgdesc='A program to control backlights (and other hardware lights)'
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+url="https://gitlab.com/dpeukert/$pkgname"
+license=('GPL-3.0-only')
+install="$pkgname.install"
+source=(
+	# $pkgrel added to make sure our cached source file doesn't get used, as the previous pkgrel used a different upstream
```
ziyao233: The renaming isn't useful: this is the first time the package is introduced to eweOS, there cannot be any type of cache. And our build-system performs no caching as well.

PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+	"$pkgname-$pkgver-$pkgrel.tar.gz::$url/-/archive/$_commit/$pkgname-$_commit.tar.gz"
```
ziyao233, line 11: Are you maintainer of the upstream repository? If so please do some type of git tagging. It's painful to keep `pkgver` synchronized with the `_commit` variable.

PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+	'fix-global-var.diff'
+	'update-manpage.diff'
```
ziyao233: Please rename them and use `.patch` as extension thus `_patch_` (mentioned below) could automatically apply them.

It's better to number them ("0001-fix-global-var.patch" "0002-update-manpage.patch", etc.). But it's not a must.

PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+            'e020deaca76f65a032f6f694f07b43af6318d46f52e0e16554a883d54c43c5519c7bbb3ecbc6e8f39d2828fbebe73d37aa82d4051b2cc652a1775286188b6ded' 
+            '36cad74b303c206035dd6941269209278c0661358820444ce97b6ef383761e258ccf56d12c80590278972db0e1be98fae3632269cbababc49690726a61fc1708')
+
+_sourcedirectory="$pkgname-$_commit"
+
+prepare() {
+	cd "$srcdir/$_sourcedirectory/"
+
+	patch --forward -p1 < "../fix-global-var.diff"
+	patch --forward -p1 < "../update-manpage.diff"
```
ziyao233: I don't think the `--foward` option is necessary[1]. If it is, then the patch requires update. Please remove the flag as we want the building procedural to be as deterministic as possible.

You could eliminate the `cd` and manually patching with `_patch_` helper[2], just like

```
prepare() {
  _patch_ "$srcdir/$_sourcedirectory"
}
```

[1]: https://github.com/eweOS/workflow/actions/runs/13284061525/job/37088684795#step:9:134
[2]: https://os-wiki.ewe.moe/dev/topic/packaging/makepkg-helpers

PKGBUILD
ziyao233: Where are these patches from? Could they be upstreamed?

The source of patches matters, please add them as comments if possible, since it helps track of the files.

PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+	./configure --prefix='/usr'
+	make
+}
+
+check() {
+	# Not using -V, as it doesn't match the real version
+	_checkoutput="$("$srcdir/$_sourcedirectory/src/$pkgname" -h)"
+	printf '%s\n' "$_checkoutput"
+	printf '%s\n' "$_checkoutput" | grep -q 'Increase brightness by value$'
+}
```
ziyao233: Suggest dropping the check function. It doesn't do any type of REAL testing actually..

PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+
+package() {
+	cd "$srcdir/$_sourcedirectory/"
+	make install DESTDIR="$pkgdir" PREFIX='/usr'
```
ziyao233, line 34: You have specified a prefix in `build()`, is it necessary here?

light.install
```diff
@@ -0,0 +1,9 @@
+post_install() {
```
ziyao233: Please drop the post-install trigger. This is the first time that the package is introduced into eweOS, thus no one would get an upgrade.

ziyao233 (changes requested):

PKGBUILD
alifnurc: I will append it to pkgver instead.

.SRCINFO
alifnurc: I see, I'v already remove them.

PKGBUILD
alifnurc: Sorry, because the repository was moved to gitlab, and is only available with commits instead of tags.

PKGBUILD
alifnurc: okey, I will remove it.

light.install
alifnurc: okey, I will remove it.

PKGBUILD
alifnurc: because it will use /usr/local instead.

PKGBUILD
alifnurc: The repo was moved to gitlab, I don't know since 'their' github account was removed.

YukariChiba (changes requested):
Please squash your commits into a single commit.

YukariChiba (changes requested):
Upstream repo is archived. See https://gitlab.com/dpeukert/light.
You may explain why this package still needs to be introduced

YukariChiba:
Looks like this PR hasn't been updated in a long time,  closing it.

PR #3146, opened 2025-02-13
[musl] 1.2.5-5: fix CVE-2025-26519 and resort patches

ziyao233:


ziyao233:
Tested on all architectures that we support.

PR #3249, opened 2025-02-23
[busybox] 1.37.0-9: Enable hwaccel for SHA-256 on AArch64

ziyao233:
This includes an under-review patch.

Reference: https://lists.busybox.net/pipermail/busybox/2025-February/091202.html


YukariChiba (approved):
Merge if you have tested it on aarch64 devices.

PR #3618, opened 2025-04-05
[bluez] 5.82-1: new upstream version

ziyao233:


ziyao233:
Seems we're hitting https://github.com/bluez/bluez/issues/726

PR #4037, opened 2025-05-14
[linux-tools] 6.14.6-1: init package

hack3ric:
Only builds some tools that 1) works with busybox, etc., 2) whose dependencies are all packaged, and 3) does not need a service script.
    
bootconfig, etc. needs GNU coreutils; cpupower, usbip, hyperv needs dinit service script; perf seems to involve so many packages; attepts to package perf will be done after this.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,144 @@
+# Maintainer: Eric Long <i@hack3r.moe>
+
+pkgbase=linux-tools
+pkgname=(bpftool tmon)
+groups=($pkgbase)
+pkgver=6.14.6
+pkgrel=1
+pkgdesc='Linux kernel tools'
+license=(GPL-2.0-only)
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url='https://www.kernel.org'
+options=(!strip !lto)
+# makedepends=(
+#   linux-headers
+#   # kernel source
+#   asciidoc xmlto
+#   # bpftool
+#   # TODO: dynamically link libbpf
+#   readline zlib libcap libelf python-docutils
+#   # tmon
+#   ncurses
+# )
+source=(https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-$pkgver.tar.xz
+        0001-bpftool-disable-bpf_jit_disasm.patch  # downstream
+        0002-turbostat-add-linux-limits.h.patch)   # will upstream
+sha256sums=('21817f1998e2230f81f7e4f605fa6fdcb040e14fa27d99c27ddb16ce749797a9'
+            'bd265ad84c3cc2e468d66582d245134cdc6c427562ae2f34ea948a4ad7d1e1d7'
+            '503043f40c2cdcf7fab1018477d81a40381338761cf48b3c15855885ba879a7f')
+
+case $CARCH in
+x86_64)
+  pkgname+=(intel-speed-select kcpuid turbostat x86_energy_perf_policy)
+  # makedepends+=(
+  #   # intel-speed-select
+  #   libnl
+  #   # turbostat
+  #   libcap
+  # )
+  ;;
+esac
+
+# HACK: list these dependencies here so that OBS can read them
+# Make sure to update both here and `makedepends=` comments above when
+# modifying dependencies
+makedepends=(linux-headers asciidoc xmlto readline zlib libcap libelf python-docutils ncurses)
+makedepends_x86_64=(libnl libcap)
+
+prepare() {
+  _patch_ linux-$pkgver
+}
+
+build() {
+  _linux="$srcdir"/linux-$pkgver
+  export CC=cc HOSTCC=cc
+
+  msg2 'bpftool'
+  cd "$_linux"/tools/bpf
+  # doesn't compile when we don't first compile bpftool in its own directory and
+  # man pages require to be also launch from the subdirectory
+  make -C bpftool all doc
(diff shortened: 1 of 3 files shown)


PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+  # kernel source
+  asciidoc xmlto
+  # bpftool
+  # TODO: dynamically link libbpf
+  readline zlib libcap libelf python-docutils
+  # tmon
+  ncurses
+)
+source=(https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-$pkgver.tar.xz
+        bpftool-disable-bpf_jit_disasm.patch)
```
ziyao233: The patch should be in form of "XXXX-*.patch"

ziyao233: There should be some explanation about upstream status of the patch (not applicable and should stay as a downstream workaround, already upstreamed IOW it's a backport, or no upstream plan yet).

PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+esac
+
+prepare() {
+  _patch_ linux-$pkgver
+}
+
+build() {
+  export CC=cc HOSTCC=cc
+
+  echo ':: bpftool'
```
ziyao233: Use `msg2` instead. This applies for other occurrences of echoes for informing building progress.

PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+  pushd linux-$pkgver/tools/bpf
```
ziyao233: I'd like to avoid these `pushd/popd` pairs. Seems `$srcdir/linux-$pkgver/xxxx` should just work.

PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+
+    echo ':: kcpuid'
+    pushd linux-$pkgver/tools/arch/x86/kcpuid
+    make
+    popd
+
+    echo ':: turbostat'
+    pushd linux-$pkgver/tools/power/x86/turbostat
+    # Fix "undeclared identifier 'PATH_MAX'"
+    sed -i '1 i\#include <linux/limits.h>' turbostat.c
```
ziyao233: Please replace `sed` with a formal patch. I guess the fix could be upstreamed (or there has been some upstreamed fixes already).

PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+    pushd linux-$pkgver/tools/power/x86/x86_energy_perf_policy
+    make
+    popd
+    ;;
+  esac
+}
+
+package_bpftool() {
+  pkgdesc='Inspection and simple manipulation of eBPF programs and maps'
+  depends=(musl readline zlib libelf libcap zstd llvm-libs binutils) # sframe not in LLVM yet
```
ziyao233: Drop `binutils`.

PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+
+  cd linux-$pkgver/tools/bpf
```
ziyao233, line 93: Please quote the variable `$pkgver`. I won't repeat the same comment for other sites where variables are expanded without quotation.

Though missing the quotation for `$pkgver` shouldn't be a problem, it's nice to keep the style aligned.

PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+  make -W runqslower_install install prefix=/usr DESTDIR="$pkgdir"
+  # fix bpftool hard written path
+  mv "$pkgdir"/usr/sbin/bpftool "$pkgdir"/usr/bin/bpftool
+  rmdir "$pkgdir"/usr/sbin
+  # install man pages
+  make -C bpftool doc-install prefix=/usr/share DESTDIR="$pkgdir"
+}
+
+package_tmon() {
+  pkgdesc='Monitoring and Testing Tool for Linux kernel thermal subsystem'
```
ziyao233: `T` in `Testing` shouldn't be uppercase.

PKGBUILD
```diff
@@ -0,0 +1,143 @@
...
+  depends=(musl ncurses)
+
+  cd linux-$pkgver/tools/thermal/tmon
+  make install INSTALL_ROOT="$pkgdir"
+}
+
+# x86_64
+
+package_intel-speed-select() {
+  pkgdesc='Intel Speed Select'
```
ziyao233: Please improve the description.

ziyao233 (changes requested):

PKGBUILD
hack3ric: It is indeed not too clear. Will state in the patch file itself.

PKGBUILD
hack3ric: Here I agree.

PKGBUILD
hack3ric: Personal taste.

PKGBUILD
```diff
@@ -0,0 +1,137 @@
...
+makedepends=(
+  linux-headers
+  # kernel source
+  asciidoc xmlto
+  # bpftool
+  # TODO: dynamically link libbpf
+  readline zlib libcap libelf python-docutils
+  # tmon
+  ncurses
+)
```
ziyao233: Not sure whether this may confuse OBS, but let's just have try (and fix it up later if necessary).

ziyao233 (dismissed):

ziyao233:
Ahhh, please update the sha256sum array before merging!

PR #4093, opened 2025-05-15
[zed] 0.186.7: init package

hack3ric:


PKGBUILD
```diff
@@ -0,0 +1,117 @@
+# Maintainer: Eric Long <i@hack3r.moe>
+
+pkgname=zed
+pkgver=0.186.7
+pkgrel=1
+_livekit_commit=3119b6ac0ef5e705b3e92630c8e558648f0892ed
+_scap_commit=08f0a01417505cc0990b9931a37e5120db92e0d0
+pkgdesc='A high-performance, multiplayer code editor from the creators of Atom and Tree-sitter'
+arch=(x86_64 aarch64 riscv64)
+url=https://zed.dev
```
ziyao233: The URL should be quoted.

PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+_git_url="https://github.com/zed-industries/zed"
```
ziyao233: This variable has been used only once in `$sources`. Consider folding the URL into `sources` directly?

PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+            'pyright: improved Python language support'
+            'rust-analyzer: improved Rust language support'
+            'org.freedesktop.secrets: to keep you logged into your Zed account')
+options=(!lto) # too much memory usage for such a big Rust project
+source=("$_git_url/archive/v$pkgver/$pkgname-$pkgver.tar.gz"
+        "git+https://github.com/hack3ric/livekit-rust-sdks.git#commit=$_livekit_commit"
+        "git+https://github.com/zed-industries/scap#commit=$_scap_commit"
+        webrtc-eweos-fixes.patch
+        scap-feature-gate-x11.patch
+        zed-disable-x11-everywhere.patch)
```
ziyao233: Please convert the patches in `XXXX-*.patch` form.

ziyao233: Please add a note about the upstream status, and also the origin of patches if applicable. Here're some examples[1][2].

[1]: https://github.com/eweOS/packages/pull/3146
[2]: https://github.com/eweOS/packages/pull/3443

PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+            '8ccef6c8740e87692024315bc8fe00b99bc389d402cd41848240a8bd10b8c6ea')
+
+_binname=zeditor
+_appid=dev.zed.Zed
+
+prepare() {
+  patch -Np1 -d livekit-rust-sdks -i ../webrtc-eweos-fixes.patch
+  patch -Np1 -d scap -i ../scap-feature-gate-x11.patch
+
+	cd $pkgname-$pkgver
```
ziyao233: There're some strange indentations, which seems to caused by mixed usage of TAB and spaces.

PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+    ;;
+  aarch64)
+    _google_arch=arm64
+    ;;
+  *)
+    _google_arch=$CARCH
+    ;;
+  esac
+
+  pushd "$srcdir/livekit-rust-sdks/webrtc-sys/libwebrtc"
```
ziyao233, line 98: Suggest replacing `pushd/popd` pairs with absolutely cds with `$srcdir` indicating the path.

PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+	sed -i '2i set -e' build_linux.sh
```
ziyao233: Is it really necessary?

PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+	cargo build --release --frozen --package zed --package cli
+}
+
+# Tests assume access to vulkan video drivers, Wayland window creation,
+# detecting system keymaps, etc. Until there is something sensical for
+# a package to test in the suite, just skip it by default.
+check() {
+	_srcenv
+	# cargo test --frozen --all-features
+}
```
ziyao233: Suggest commenting the whole function away.

webrtc-eweos-fixes.patch
ziyao233, line 1: Please split the patch: provide the libatomic and musl parts as plain files instead of patches.

webrtc-eweos-fixes.patch
```diff
@@ -0,0 +1,69 @@
...
++     defines = [ "OS_CHROMEOS" ]
++   }
++-
++-  if ((!is_chromeos || default_toolchain != "//build/toolchain/cros:target") &&
++-      (!use_custom_libcxx || current_cpu == "mipsel")) {
++-    libs = [ "atomic" ]
++-  }
++ }
++ 
++ config("libcap") {
```
ziyao233: Please explain about the patch. Same for the next one.

PKGBUILD
```diff
@@ -0,0 +1,117 @@
...
+
+	cd $pkgname-$pkgver
+
+  patch -Np1 -i ../zed-disable-x11-everywhere.patch
+  echo -e "\n[patch.'https://github.com/zed-industries/scap']\nscap = { path = '$srcdir/scap' }" >> Cargo.toml
+  cargo update -p scap
+	cargo fetch --locked --target $RUSTHOST
+
+  # even thin LTO consumes too much memory
+  sed -i '/lto = "thin"/ d' Cargo.toml
```
ziyao233: Please replace the sed with a formal patch.

ziyao233 (changes requested):

PKGBUILD
hack3ric: They are not applied to the same directory, so no

PKGBUILD
hack3ric: Yes. The script does not do this.

PKGBUILD
hack3ric: The description should be better done in patch file itself.

PKGBUILD
hack3ric: I’d agree in other contexts, but not here. It looks better with pushd/popd since there’s no need to write the first cd again.

webrtc-eweos-fixes.patch
hack3ric: File name and patch content should be self-explanatory.

PKGBUILD
ziyao233: I'd prefer to keep a downstream/upstream/not upstreamed yet tag above sources, which provides a better overview at least to me. Anyway I'm fine with both ways generally, as long as the origin is tracked and there're notes of reasons for patching.

PKGBUILD
hack3ric: Not too big of a deal

PKGBUILD
ziyao233: I prefer to be explicit in any of the contexts: one'll spend more of their time in tracking the $PWD where operations happen than maintaining one additional "cd" command.

I won't insist on this detail though.

ziyao233 (approved):

PR #4127, opened 2025-05-17
[tinymist] 0.13.12-1: init package

hack3ric:


hack3ric:
CI will fail due to missing `cargo-insta`. Should be merged after it.

PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+
+prepare() {
+  cd $pkgname-$pkgver
+  cargo fetch --locked --target $RUSTHOST
+}
+
+build() {
+  cd $pkgname-$pkgver
+  export OPENSSL_NO_VENDOR=true
+  cargo build --frozen --release
```
ziyao233: Suggestion: I'm not sure whether Cargo takes `--all-features` as default. If not, always adding the option will save some work when some new features are introduced in the future (there seems to be no currently).

PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+}
+
+check() {
+  cd $pkgname-$pkgver
+  export OPENSSL_NO_VENDOR=true
+  cargo test --frozen -- --skip e2e
```
ziyao233: Could you please provide some information on the reason for skipping the test? Please add a note to the PKGBUILD as well if applicable.

hack3ric: The E2E test is run at the next line as it requires some setups done in the script.

ziyao233: Oops, my mistake, thanks.

ziyao233:
Recheck after #4126

PR #2995, opened 2025-01-17
[busybox] 1.37.0-8: fix generating incorrect interface names

JulianDroske:
This is an alternative to PR #2983, since we haven't figure out how to correctly trigger NM so far.

* Revert: [fix mdev helper bug with iwlwifi mac address](https://github.com/JulianDroske/eweos-packages/commit/9806fe4780a7fc56e5a8838c6e1a171b171328a4)
* The script will try to use permanent mac address first, if possible.

PR #4128, opened 2025-05-17
[libudev] 1.0.3.r5.bbeb7ad-3: scan all network interfaces

JulianDroske:
It makes NetworkManager think all network interfaces are ready, and will manage all of them instead of showing "unmanaged" states.

After this PR, the patch force-platform-init.patch [1] [2] may no longer needed.

PS: It's still recommended to install "ethtool" (to avoid conflicts [3]).

[1]: https://github.com/illiliti/libudev-zero/issues/72
[2]: https://github.com/JulianDroske/eweos-packages/blob/9eceb0d729f6677c3c1d71a8c30e0f610c5310b2/force-platform-init.patch
[3]: https://github.com/eweOS/packages/pull/2995


ziyao233:
Thanks for the PR, but

> It makes NetworkManager think all network interfaces are ready, and will manage all of them instead of showing "unmanaged" states.

This may cause race conditions: our mdevd scripts may adjust interface names, thus it isn't safe to consider the interface ready (settled) before the script completes, or programs like NetworkManager may operates on an interface whose hasn't been correctly assigned.

The keypoint is that libudev-zero isn't bound to any specific devd implementation, thus it's somehow hard to notify the consumer programs (e.g. NetworkManager) about state of the device. Note that simply sending a message through libudev-helper won't work, as it's not guaranteed the consumer program has already been running when the device is settled.

It's required to work out a better solution for device states, let's close the PR for now to avoid future confusion.

PR #4311, opened 2025-06-05
[kmod] 34.2-1: new upstream version, disable test-depmod

JulianDroske:
Changes:

1. The compiling method is switched from Makefile to Meson (thanks to @hack3ric 's work [1]).
2. The test test-depmod has been disabled by patching meson.build as which would always fail *(in depmod_modules_order_for_compressed and depmod_modules_outdir)* under LLVM/clang (see below).
3. Previous patches (lfs64.patch fix-test-ldflags.patch) may not be needed, so dropped.

**About Change 2**
- Caused by: Incorrect order generated in modules.alias.
- Caused by: Different orders of symbols in final .o/.ko files
- Caused by: Clang does not implement reverse symbol orders like that in gcc [2].

It should not make a difference to regular use, while the problem is from clang instead, so it's disabled to pass the check.

---

Based on and replaces https://github.com/eweOS/packages/pull/4129

Reverts 7e46ea3

[1]: https://github.com/eweOS/packages/commit/782b64235b3183975ec2a67ea584d5de88a3e6dc
[2]: https://gcc.gnu.org/pipermail/gcc-help/2019-March/136880.html


PKGBUILD
```diff
@@ -1,95 +1,79 @@
 # Maintainer: Yukari Chiba <i@0x7f.cc>
 # Contributor: Aleksana QwQ <me@aleksana.moe>
+# Contributor: Julian Droske <juliandroske@outlook.com>
 
 pkgbase=kmod
 pkgname=(kmod libkmod)
-pkgver=33
-pkgrel=2
+pkgver=34.2
```
ziyao233, line 7: Please fix the prefix of the first commit, or just squash these two together.

PKGBUILD
```diff
@@ -1,95 +1,79 @@
...
 
-prepare()
-{
-  _patch_ "$pkgbase-$pkgver"
+prepare() {
+  cd "${pkgbase}-${pkgver}"
+  # Omit -r, which is implemented using symlinks in package()
+  sed -i 's/ln -sfr/ln -sf/g' meson.build
+  # Disable test test-depmod which would always fail using LLVM/clang for now
+  sed -i '/test-depmod/d' testsuite/meson.build
```
ziyao233: Please create patches for each purpose. Sed may interact unexpectedly with file content.

PKGBUILD
```diff
@@ -1,95 +1,79 @@
...
+  pkgdesc="${pkgdesc} library"
+  license=(LGPL-2.0-or-later)
+  provides=(libkmod.so)
 
-  install -d $pkgdir/usr
-  mv libkmod/* $pkgdir/usr
+  cd build
+  install -d "${pkgdir}/usr/lib"
+  rm -rf libkmod.so.*.p
+  mv libkmod.so* "${pkgdir}/usr/lib/"
```
ziyao233: You've `_pick_`'ed libkmod-related files in `package_kmod()`, why not just install the picked file?

ziyao233 (changes requested):
The changes mostly go in the right way, there're only two problems to consider,

- You need to fix the CI: IOW, override the testsuite logic of looking for kernel's source directory to use eweOS kernel always. It's totally reasonable for a builder not to run a eweOS kernel.
- Could you keep Eric's credit on the patch? Please at least add a Co-developed-by tag (or Co-authored-by) in the commit message.

PKGBUILD
JulianDroske: I misunderstood the `_pick_` function so did the wrong thing, while the usage of `_pick_` in `package_kmod()` was incorrect. They will be fixed in the next commit push.

PKGBUILD
```diff
@@ -1,95 +1,85 @@
...
-
-  local kver kdir=${kdirs[0]%/Makefile}
-  IFS=/ read _ _ _ kver _ <<<"$kdir"
-  CFLAGS+=" -fuse-ld=lld"
-  # use libtool to avoid symlink failure
-  export MAKEFLAGS="CC=cc CXX=c++ -j$JOBS"
-  # FIXME: fix errors in test
-  make -C "$pkgname-$pkgver" check KDIR="$kdir" KVER="$kver" LLVM=1 LLVM_IAS=1 || :
+  # Use the reasonable kernel source instead of the exact running one
+  MODVER=$(pacman -Qi linux-devel |awk -e '{ if ($1 == "Version") { print ($3) } }')
```
ziyao233:
```suggestion
  MODVER=$(pacman -Qi linux-devel | awk -e '{ if ($1 == "Version") { print ($3) } }')
```

PKGBUILD
```diff
@@ -1,95 +1,85 @@
...
+makedepends=(meson scdoc pahole) # pahole is used in test (at least when using LLVM)
+checkdepends=(linux-devel libelf lld linux)
+source=("https://www.kernel.org/pub/linux/utils/kernel/${pkgname}/${pkgname}-${pkgver}.tar.xz"
+        depmod-search.conf
+        depmod.hook
+        depmod.script
+        # Omit -r, which is implemented using symlinks in package_kmod()
+        avoid-unsupported-ln-relative.patch
+        # Disable test test-depmod which would always fail using LLVM/clang
+        # for now
```
ziyao233: Patches should be named like `0001-avoid-unsupported-ln-relative.patch`

PKGBUILD
```diff
@@ -1,95 +1,85 @@
...
-            '540e07c6c5623d7b519a6f9bb18e151c2f764c1d77c22c842d09f2d3daafb844'
+license=(LGPL-2.1-or-later GPL-2.0-or-later)
+depends=(musl openssl xz zlib zstd) # shared between kmod and libkmod
+makedepends=(meson scdoc pahole) # pahole is used in test (at least when using LLVM)
+checkdepends=(linux-devel libelf lld linux)
+source=("https://www.kernel.org/pub/linux/utils/kernel/${pkgname}/${pkgname}-${pkgver}.tar.xz"
+        depmod-search.conf
+        depmod.hook
+        depmod.script
+        # Omit -r, which is implemented using symlinks in package_kmod()
```
ziyao233: OBS is bad at parsing arrays, thus it's better to move these comments out of the array. It should be fine to put them above `source=`.

PKGBUILD
ziyao233: Could you please describe the reason of failures here as well? Copying part of the commit message here should be okay, which helps future readers to figure out why.

ziyao233 (changes requested):

ziyao233 (approved):
Looks good to me now! Will give it a test tomorrow and merge then if everything goes smoothly.

ziyao233:
Works on my machine, thanks!

PR #4380, opened 2025-06-10
[wlroots0.19] 0.19.0-1: init package

ziyao233:


PKGBUILD
```diff
@@ -0,0 +1,49 @@
+# Maintainer: Yukari Chiba <i@0x7f.cc>
```
hack3ric: Please set the maintainer to yourself

PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+        -Dbackends=drm,libinput \
+        -Drenderers=gles2 \
+        -Dexamples=false \
+        -Dxwayland=disabled \
+        -Dxcb-errors=disabled
+    ninja -C build
+}
+
+package_wlroots0.19-devel() {
+    depends=(wlroots0.19=$pkgver)
```
hack3ric: Since other fields are quoted, please also wrap this with quotation marks.

hack3ric (dismissed):

PKGBUILD
ziyao233: Thanks. Fixed some other unquoted variable expansions as well.

hack3ric (approved):

PR #4386, opened 2025-06-11
[lxc] 6.0.4-1: init package

who53:


who53:
Should I add myself as  a contributor instead

who53:
v2: remove perl and python from depends
perl not needed
python only needed for generating apparmor rules which we dont support

ziyao233:
> Should I add myself as a contributor instead

It depends on you. eweOS don't have a lot of active members and we take care of others' packages from time to time. You're welcome if you want to be listed as maintainer and may be Cc'ed on future changes of the package.

PKGBUILD
```diff
@@ -0,0 +1,34 @@
+# Maintainer: Deepak Meena <notwho53@gmail.com>
+
+pkgname=lxc
+pkgver=6.0.4
+pkgrel=1
+pkgdesc="Linux Containers"
```
ziyao233: Please improve the description, here's a general documentation on writing package descriptions[1].

Suggest "The Linux Containers tool set for kernel-level isolation"

[1]: https://os-wiki.ewe.moe/en/dev/topic/packaging/pkgdesc-guideline

PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url="https://linuxcontainers.org"
+depends=('bash' 'libseccomp' 'libcap' 'rsync' 'wget')
```
ziyao233: I think the package depends on libc (`musl`) as well since it contains binaries.

who53:
> > Should I add myself as a contributor instead
> 
> It depends on you. eweOS don't have a lot of active members and we take care of others' packages from time to time. You're welcome if you want to be listed as maintainer and may be Cc'ed on future changes of the package.

Ok then I will keep it as it is now :D

PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+makedepends=('meson' 'lua' 'python-setuptools' 'dbus')
```
ziyao233: Are you sure dbus is a build-time dependency instead of a runtime one? It looks a little strange.

PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+license=('LGPL')
```
ziyao233: Please use SPDX license identifiers in the license entry. For LXC, it should be `LGPL-2.0-or-later`[1]

[1]: https://github.com/lxc/lxc/blob/76975058b99252de6f1cfbbc9d3dba9029365287/lxc.spec.in#L49

PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+	'etc/default/lxc')
+source=("https://linuxcontainers.org/downloads/lxc/$pkgname-${pkgver}.tar.gz")
+sha256sums=('872d26ce8512b9f993d194816e336bf9f3ad8326f22dc24ef0f01f85599fa8b9')
+
+
+build() {
+  export LDFLAGS="$LDFLAGS -lunwind"
+
+  ewe-meson $pkgbase-$pkgver build -Dapparmor=false \
+      -Dman=false \
```
ziyao233, line 23: Could you please add a note about the reason that man-pages aren't built? Something like

```
# Disable man-pages for now since dockbook2x is missing
```

will be helpful for future maintenance.

ziyao233 (changes requested):
Though there're some minor points to be fixed, the PKGBUILD mostly looks good to me.

One additional question for avoiding duplicated work: do you have a plan to write a dinit service for lxc, to allow users directly start a container as system-level service? It isn't necessary but is something nice to have, and I could do this after lxc is basically packaged.

PKGBUILD
who53: yes it is but looking closely its tied with systemd so dropping it

who53:
I have made changed required. Please review again :)

who53:
> Though there're some minor points to be fixed, the PKGBUILD mostly looks good to me.
> 
> One additional question for avoiding duplicated work: do you have a plan to write a dinit service for lxc, to allow users directly start a container as system-level service? It isn't necessary but is something nice to have, and I could do this after lxc is basically packaged.

sorry but it would be great if you do it, I am not yet familiar with dinit

ziyao233 (dismissed):
lgtm, will test and merge it later if everything goes smoothly :)

ziyao233:
> > Though there're some minor points to be fixed, the PKGBUILD mostly looks good to me.
> > One additional question for avoiding duplicated work: do you have a plan to write a dinit service for lxc, to allow users directly start a container as system-level service? It isn't necessary but is something nice to have, and I could do this after lxc is basically packaged.
> 
> sorry but it would be great if you do it, I am not yet familiar with dinit

Sure, that's okay

PKGBUILD
ziyao233: > yes it is but looking closely its tied with systemd so dropping it

looking through the CI failures, I think you should move it to depends, thanks!

who53: no that was skill issue :P fixed now 

 it was ewe-meson not playing well with my comment

hack3ric:
Tested on my setup and it works with basic setup. Merging.

hack3ric (approved):

PR #4396, opened 2025-06-14
[limine] 9.3.3-2: add hook to auto update limine.conf

YukariChiba:


limine-mkconfig
```diff
@@ -1,12 +1,12 @@
...
 
-exec >/dev/stderr
+exec >&2
 
 source /etc/os-release
 source /etc/limine.conf
 
 OPTSTRING=":o:b:"
-CFGOUT=/dev/stdout
+CFGOUT="&1"
```
hack3ric, line 7: This seems to not work properly (even before the changes): output configuration will still be printed to stderr since the `exec` basically points stdout file to stderr. I'd suggest replacing all `echo` log below with `>&2 echo` and remove the `exec` at the top of the file.

hack3ric (changes requested):

PR #4413, opened 2025-06-15
[tinymist] 0.13.12-3: remove --all-features, fix LTO and tests

hack3ric:


ziyao233 (dismissed):
With some offlist discussion, we agreed it's unreasonable not to allow testing as root user. So the patch should be marked as to be upstreamed.

With this fixed, this PR looks good to me.

hack3ric:
For the record, the patch is upstreamed as https://github.com/Myriad-Dreamin/tinymist/pull/1817.

PR #4414, opened 2025-06-16
[rust] 1.87.0-2: remove rls, add rust-analyzer-proc-macro-srv, provide rustdoc

hack3ric:
rust-analyzer-proc-macro-srv is required for rust-analyzer. It is shipped along with rust package because it strongly depends on current rustc version. rls is deprecated in favor of rust-analyzer.

Also includes a PKGBUILD style fix commit.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -3,7 +3,7 @@
 pkgbase=rust
 pkgname=(rust rust-src)
 pkgver=1.87.0
-pkgrel=1
+pkgrel=2
 pkgdesc="Systems programming language focused on safety, speed and concurrency"
 arch=(x86_64 aarch64 riscv64 loongarch64)
 url='https://www.rust-lang.org/'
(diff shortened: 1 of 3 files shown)


PKGBUILD
```diff
@@ -3,67 +3,57 @@
...
-  0003-drop-latomic-for-riscv-targets.patch
-)
+depends=(musl llvm-libs musl-static curl libssh2 openssl)
+makedepends=(rust llvm-devel libffi perl python cmake ninja)
+source=(https://static.rust-lang.org/dist/rustc-$pkgver-src.tar.gz
+        config.toml.tpl
+        0001-musl-static.patch
+        0002-disable-no-default-libraries.patch
+        0003-drop-latomic-for-riscv-targets.patch
+        0004-bootstrap-Change-libexec-dir.patch)  # install rust-analyzer-proc-macro-srv in /usr/lib rather than /usr/libexec
```
ziyao233, line 19: It's better to move the comment out of the array to avoid break OBS :(

ziyao233 (approved):

PKGBUILD
ziyao233: Nevermind, comments out of the array are okay.

PR #4416, opened 2025-06-16
[llvm] 20.1.7-2: enable clang-tools-extra, package clang-tools

hack3ric:
Includes cleanup commits.


PKGBUILD
```diff
@@ -44,40 +40,77 @@ sha256sums=('91865189d0ca30ca81b7f7af637aca745b6eeeba97c5dfb0ab7d79a1d9659289'
...
+  "usr/share/clang/clang-format*"
 )
 
 FLIST_clang=(
-  "usr/bin/*clang*"
-  "usr/bin/c-index-test"
   "usr/bin/cc"
   "usr/bin/c++"
+  "usr/bin/clang"{,-20}
+  "usr/bin/clang++"{,-20}
```
ziyao233: I don't think we should hardcode the version here, maybe define `_majorver` as `${pkgver%%.*}"` and use `"usr/bin/clang"{,-"$majorver"}`?

ziyao233: This applies for flang as well.

ziyao233 (changes requested):

ziyao233 (approved):
Tested on loongarch64.

PR #4682, opened 2025-06-29
[prjtrellis] 1.4-1: init package

ziyao233:


PKGBUILD
```diff
@@ -0,0 +1,39 @@
+# Maintainer: Yao Zi <ziyao@disroot.org>
+
+pkgname=prjtrellis
+pkgver=1.4
+_dbcommit=4dda149b9e4f1753ebc8b011ece2fe794be1281a # 2023.09.28
+pkgrel=1
+pkgdesc='Libraries and utilities for open-source toolchain of Lattice ECP5/MachXO2 series FPGAs'
+url='https://archlinux.org/packages/prjtrellis'
```
YukariChiba: Why use arch package info page as `url`?

YukariChiba (changes requested):

PKGBUILD
ziyao233: Ooops, it's a mistake when pasting the URL. Fixed in the new HEAD.

PR #4920, opened 2025-08-06
[chez-scheme] 10.2.0-1: init package

rmolives:


PKGBUILD
```diff
@@ -0,0 +1,36 @@
+# Maintainer: RMOlive <rmolives@wumoe.org>
+
+pkgname=chez-scheme
+pkgver=10.2.0
+pkgrel=1
+pkgdesc="Compiler and run-time system for the language of the Revised^6 Report on Scheme (R6RS), with numerous extensions."
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+url="https://github.com/cisco/ChezScheme"
+depends=()
```
ziyao233, line 9: This empty depends array looks suspicious and is overwritten later. Please remove it.

PKGBUILD
```diff
@@ -0,0 +1,36 @@
+# Maintainer: RMOlive <rmolives@wumoe.org>
+
+pkgname=chez-scheme
+pkgver=10.2.0
+pkgrel=1
+pkgdesc="Compiler and run-time system for the language of the Revised^6 Report on Scheme (R6RS), with numerous extensions."
```
ziyao233, line 6: Please improvement the pkgdesc, suggest something like

```
Compiler and runtime for R6RS release of Scheme
```

Reference: https://os-wiki.ewe.moe/dev/topic/packaging/pkgdesc-guideline

PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+url="https://github.com/cisco/ChezScheme"
+depends=()
+license=('Apache-2.0')
+depends=('ncurses' 'libutil-linux')
```
ziyao233, line 11: Please list out all the direct dependencies to the package. At leaset musl is missed here.

PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+conflicts=('petite-chez-scheme' 'chez-scheme-git')
+source=(
+    "https://github.com/cisco/ChezScheme/releases/download/v$pkgver/csv$pkgver.tar.gz"
+    "clearcache.patch"
```
ziyao233, line 15: Similarly to the `racket` package: what is the patch for and what's its origin? Is it appropriate for upstreaming if it isn't backported from mainline?

Please add a comment for explaining. You could take [firefox](https://github.com/eweOS/packages/blob/5c9332ee75052d025a4ef87c47ad54eaf87f3634/PKGBUILD#L40-L49) and [nodejs](https://github.com/eweOS/packages/blob/c57ce80ef6b1ec8eb7881b5cf6ed9b568faa42f4/PKGBUILD#L13-L18) for reference.

PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+)
+_archivename=csv$pkgver
+
+prepare() {
+  _patch_ "${_archivename}"
+}
+
+build() {
+  cd "$srcdir/${_archivename}"
+  ./configure --installprefix=/usr --temproot=$pkgdir --threads --installschemename=chez --installscriptname=chez-script --disable-x11 CC=clang
```
ziyao233, line 29: Is it really necessary to specify `CC=clang` explicitly when running configure? The default makepkg.conf should have already exported `CC=clang` as an environment variable.

ziyao233 (changes requested):

PR #4921, opened 2025-08-06
[racket] 8.17-1: init package

rmolives:


PKGBUILD
```diff
@@ -0,0 +1,46 @@
+# Maintainer: RMOlive <rmolives@wumoe.org>
+
+pkgbase=racket
```
ziyao233, line 3: pkgbase isn't necessary when pkgname isn't an array. Please drop it.

PKGBUILD
```diff
@@ -0,0 +1,46 @@
...
+pkgrel=1
+pkgdesc='A full-spectrum language with DrRacket IDE'
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+url='https://racket-lang.org/'
+license=('Apache-2.0' 'GPL-3.0-only' 'LGPL-3.0-only' 'custom')
+depends=('gtk3')
+makedepends=('gsfonts' 'sqlite')
+options=('!strip' '!emptydirs')
+conflicts=('racket-docs')
+replaces=('racket-docs')
```
ziyao233, line 15: eweOS doesn't even have this package. Please drop conflicts and replaces.

PKGBUILD
```diff
@@ -0,0 +1,46 @@
...
+
+pkgbase=racket
+pkgname='racket'
+pkgver=8.17
+pkgrel=1
+pkgdesc='A full-spectrum language with DrRacket IDE'
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+url='https://racket-lang.org/'
+license=('Apache-2.0' 'GPL-3.0-only' 'LGPL-3.0-only' 'custom')
+depends=('gtk3')
```
ziyao233, line 11: Please lay out all the direct dependencies: at least libc is necessary, isn't it?

PKGBUILD
```diff
@@ -0,0 +1,46 @@
...
+
+prepare() {
+  _patch_ "$pkgname-$pkgver"
+  cd "$pkgbase-$pkgver"
+  echo "Icon=drracket" >> share/pkgs/drracket/drracket/drracket.desktop
+}
+
+build() {
+  cd "$pkgbase-$pkgver/src"
+  [ "$CARCH" == "x86_64" ] && export CFLAGS+=" -fPIC -ffat-lto-objects"
```
ziyao233, line 33: Why fat LTO objects is and only is necessary for x86_64 platforms? Please add a comment to explain it.

PKGBUILD
```diff
@@ -0,0 +1,46 @@
+# Maintainer: RMOlive <rmolives@wumoe.org>
+
+pkgbase=racket
+pkgname='racket'
+pkgver=8.17
+pkgrel=1
+pkgdesc='A full-spectrum language with DrRacket IDE'
+arch=('x86_64' 'aarch64' 'riscv64' 'loongarch64')
+url='https://racket-lang.org/'
+license=('Apache-2.0' 'GPL-3.0-only' 'LGPL-3.0-only' 'custom')
```
ziyao233, line 10: As specified in `src/LICENSE.txt`, racket is distributed under `Apache-2.0 AND MIT`. MIT license is missed in the licenses array.

Furthermore, I don't see any components using GPL-3.0-only license. And all LGPL components that ***may*** be statically linked (I didn't check precisely) allow usage of later versions, so `LGPL-3.0-only` should be `LGPL-3.0-or-later`.

For the `custom` license, please provide its name as well. I guess it should be `custom:libscheme` but please double check my finding.

`LICENSE.txt` also states that `random.inc`, used by Racket "BC" executables, is under BSD-3-Clause. Does this PKGBUILD ship the "BC" executable? If so, BSD-3-Clause should be listed in licenses as well.

clearcache.patch
```diff
@@ -0,0 +1,11 @@
...
+--- a/src/ChezScheme/c/clearcache.c	2025-04-11 07:46:18.000000000 +0800
++++ b/src/ChezScheme/c/clearcache.c	2025-07-30 17:14:37.496107332 +0800
+@@ -45,6 +45,7 @@
+ #ifdef S_TARGET_OS_IPHONE
+   sys_icache_invalidate((void *)start, (char *)end-(char *)start);
+ #else
++  extern void __clear_cache(void*, void*);
+   __clear_cache((char *)start, (char *)end);
+ # if defined(__clang__) && defined(__aarch64__) && !defined(__APPLE__)
+   /* Seem to need an extra combination of barriers here to make up for
```
ziyao233, line 11: What is this patch for and what's its origin? Could you please add a comment above the `sources` array in PKGBUILD to explain it?

Furthermore, if it isn't a backport, is this patch appropriate for upstreaming?

ziyao233 (changes requested):

PR #5093, opened 2025-08-14
[cmake] 4.1.0-1: new upstream version

YukariChiba:


ziyao233 (approved):
Tested with qt6-base and openal-soft, packages are installed to the correct directory.

PR #5260, opened 2025-08-24
[kicad] 9.0.4-1: init package

ziyao233:


ziyao233:
Recheck after #5251, #5252, #5255, #5256, #5259

PR #5600, opened 2025-10-13
[wireplumber] 0.5.12-1: new upstream version

YukariChiba:


ziyao233 (approved):
Nothing seems to break on my machine with the upgrade

Issue #5581, opened 2025-10-12
Replace p7zip with upstream 7zip


JulianDroske:
Since several CVEs have been uncovered in older versions of 7zip recently [^1], specifically:

- [CVE-2025-55188](https://nvd.nist.gov/vuln/detail/CVE-2025-55188)
- [CVE-2025-53817](https://nvd.nist.gov/vuln/detail/CVE-2025-53817)
- [CVE-2025-53816](https://nvd.nist.gov/vuln/detail/CVE-2025-53816)

p7zip, which is derived from 7z and have been inactive without a release for two years [^2], could not be considered to be safe. To get rid of such security issues, the upstream `7zip` is taken into account, which is more actively maintained while fixes all these exploits in the latest release.

## Steps
- [x] Add package `7zip` conflicting p7zip #5582, #5605
  - [ ] Add missing man pages?
- [x] Remove package `p7zip`
  - [x] Check for dependents

[^1]: https://nvd.nist.gov/vuln/search#/nvd/home?cpeFilterMode=cpe&cpeName=cpe:2.3:a:7-zip:7-zip:-:*:*:*:*:*:*:*&resultType=records
[^2]: https://github.com/p7zip-project/p7zip/releases/tag/v17.05

PR #5605, opened 2025-10-13
[7zip] 25.01-1: init package

JulianDroske:
This package `conflicts` and `provides` `p7zip` to replace it because of #5581 

Resulting structure:

`/usr/lib/7zip/`: `7z.so`
`/usr/bin/`: `7z`, `7za`, `7zr`
`/usr/share/doc/7zip/`: `DOC/`, documentations
`/usr/share/licenses/7zip/`: two license files symlinking to those under `DOC/`

Upstream `7zip` does not provide any man page.

There's no executable `7z*` under `/usr/lib/7zip` since we got rid of shell script wrappers used by other distros and p7zip, by modifying the library search path, shown in the patch file. It actually organize files in a more clear way.

`/usr/lib/7zip/` is used by `7z` for searching for external encoders/decoders so keep it.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,0 +1,77 @@
+# Maintainer: Julian Droske <juliandroske@outlook.com>
+
+pkgname=7zip
+pkgver=25.01
+pkgrel=1
+pkgdesc="Command-line file archiver with high compression ratio"
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url="https://7-zip.org"
+license=('LGPL-2.1-or-later' 'BSD-3-Clause' 'custom:unRAR')
+provides=(p7zip)
+conflicts=(p7zip)
+replaces=(p7zip)
+depends=(
+  musl
+  llvm-libs
+)
+# 0001-adjust-7z.so-load-path-for-eweos.patch
+# 	The executable `7z` searches for its requistie `7z.so` from the same
+# 	directory (intended to store plugins) as where it lies at runtime.
+# 	This causes problems when `7z` is installed under `/usr/bin/` as a
+# 	command while `7z.so` is put in `/usr/lib/7zip/` as FHS implies.
+# 	This patch modifies the search path so that 7z always search for plugins
+# 	(including 7z.so) under `/usr/lib/7zip` without any shell script wrappers.
+source=(
+  'https://www.7-zip.org/a/7z2501-src.tar.xz'
+  '0001-adjust-7z.so-load-path-for-eweos.patch'
+)
+sha256sums=(
+  'ed087f83ee789c1ea5f39c464c55a5c9d4008deb0efe900814f2df262b82c36e'
+  '9d3604f51ddcdee78631c31d90e93949cb41772a91f6d9b98b8fac5dc90a028a'
+)
+
+prepare(){
+  _patch_ .
+}
+
+build(){
+  cd "${srcdir}"
+  export O=build
+
+  # 7z - main program that uses plugins
+  make -C CPP/7zip/UI/Console -f makefile.gcc
+  # 7z.so - necessary library required by 7z
+  make -C CPP/7zip/Bundles/Format7zF -f makefile.gcc
+  # 7za - more lightweight
+  make -C CPP/7zip/Bundles/Alone -f makefile.gcc
+  # 7zr - only handle .7z archives
+  make -C CPP/7zip/Bundles/Alone7z -f makefile.gcc
+}
+
+package(){
+  cd "${srcdir}"
+
+  # target directories
+  install -dm0755 \
+  	"${pkgdir}/usr/bin" \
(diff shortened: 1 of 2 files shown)


PKGBUILD
```diff
@@ -0,0 +1,69 @@
...
+pkgname=7zip
+pkgver=25.01
+pkgrel=1
+pkgdesc="Command-line file archiver with high compression ratio"
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url="https://7-zip.org"
+license=('LGPL-2.1-or-later' 'BSD-3-Clause' 'custom:unRAR')
+provides=(p7zip)
+conflicts=(p7zip)
+depends=(
```
ziyao233, line 13: You should add `replaces=(p7zip)` as wel :)

PKGBUILD
```diff
@@ -0,0 +1,69 @@
...
+
+package(){
+  cd "${srcdir}"
+
+  # target directories
+  install -dm0644 \
+    "${pkgdir}/usr/bin" \
+    "${pkgdir}/usr/lib/7zip" \
+    "${pkgdir}/usr/share/doc/7zip" \
+    "${pkgdir}/usr/share/licenses/7zip"
```
ziyao233: pacman complains when installing the package,

```
warning: directory permissions differ on /usr/bin/
filesystem: 755  package: 644
```

It seems wrong to use `0644` on directories, and `0755` is more appropriate.

PKGBUILD
```diff
@@ -0,0 +1,69 @@
...
+  install -Dm0755 \
+    CPP/7zip/Bundles/Format7zF/build/7z.so \
+    "${pkgdir}/usr/lib/7zip"
+
+  # Docs & licenses
+  cp -a DOC "${pkgdir}/usr/share/doc/7zip/"
+  ln -s /usr/share/doc/7zip/DOC/{License,unRarLicense}.txt \
+    "${pkgdir}/usr/share/licenses/7zip/"
+}
+
```
ziyao233: Please remove this extra empty line :)

PKGBUILD
```diff
@@ -0,0 +1,69 @@
...
+url="https://7-zip.org"
+license=('LGPL-2.1-or-later' 'BSD-3-Clause' 'custom:unRAR')
+provides=(p7zip)
+conflicts=(p7zip)
+depends=(
+  musl
+)
+source=(
+  'https://www.7-zip.org/a/7z2501-src.tar.xz'
+  '0001-adjust-7z.so-load-path-for-eweos.patch'
```
ziyao233, line 26: It'll be better if you could add a comment to explain the patch's purpose, e.g.,

> 7z searches its shared library, 7z.so, from the same directory that contains the executable, and loads it at runtime. However, this breaks if we install executables in `/usr/bin` but keep the library in `/usr/lib/7zip`. Hack the library-searching logic to account for the library/executable separated case.

ziyao233 (changes requested):

ziyao233:
Forgot to mention it: you should include `llvm-libs` in `depends` array as well, since this package makes use of C++ runtime libraries like `libunwind.so` and `libc++.so`.

PKGBUILD
```diff
@@ -0,0 +1,77 @@
...
+
+package(){
+  cd "${srcdir}"
+
+  # target directories
+  install -dm0755 "${pkgdir}/usr/bin"
+  install -dm0644 \
+    "${pkgdir}/usr/lib/7zip" \
+    "${pkgdir}/usr/share/doc/7zip" \
+    "${pkgdir}/usr/share/licenses/7zip"
```
ziyao233: All these directories should be installed as `0755`, or non-owner users (basically means non-root ones) cannot descend into the directories after installing the package. This also breaks 7zip itself,

```
$ 7z x ~/Downloads/Telegram\ Desktop/MP020-r0p5-50rel0-37522aa9-0.tar.gz 

7-Zip 25.01 (x64) : Copyright (c) 1999-2025 Igor Pavlov : 2025-08-03
 64-bit locale=C.UTF-8 Threads:4 OPEN_MAX:1024


Codec Load Error: /usr/lib/7zip/7z.so : errno=13 : Permission denied

Scanning the drive for archives:
1 file, 2016896 bytes (1970 KiB)                 

Extracting archive: /home/ziyao/Downloads/Telegram Desktop/MP020-r0p5-50rel0-37522aa9-0.tar.gz
ERROR: /home/ziyao/Downloads/Telegram Desktop/MP020-r0p5-50rel0-37522aa9-0.tar.gz
Cannot open the file as archive

    
Can't open as archive: 1
Files: 0
Size:       0
Compressed: 0
```

`chmod 0755 /usr/lib/7zip` just fixes the issue.

ziyao233 (changes requested):

ziyao233 (approved):

PR #5723, opened 2025-10-24
[gtk4] 4.20.2-2: enable vulkan and cups

YukariChiba:


ziyao233 (approved):
Tested on my machine, by inspecting strace I could confirm Vulkan works with the patch. Nothing seems to break.

PR #6058, opened 2025-11-23
[font-noto-emoji] 2.048-2: use SPDX and simplify license installation

ziyao233:


PKGBUILD
YukariChiba: bump pkgrel

YukariChiba (changes requested):

YukariChiba (approved):

PR #6139, opened 2025-11-27
[turnstile] 0.1.11-1: new upstream version and reformat

WeirdGumi:


PKGBUILD
```diff
@@ -1,36 +1,40 @@
 # Maintainer: Yukari Chiba <i@0x7f.cc>
 
 pkgname=turnstile
-pkgver=0.1.10
-pkgrel=3
+pkgver=0.1.11
+pkgrel=1
 pkgdesc='Independent session/login tracker'
-url='https://github.com/chimera-linux/turnstile'
+url=https://github.com/chimera-linux/turnstile
```
ziyao233: Please keep the quote. Dropping it is okay in the case, but URL contains some characters that may not be considered safe at the first glance, causing unnecessary confusion, however keeping it as-is is always safe.

PKGBUILD
```diff
@@ -1,36 +1,40 @@
...
 arch=(x86_64 aarch64 riscv64 loongarch64)
 license=(BSD-2-Clause)
 depends=(pam)
```
ziyao233: This probably depends on `libc.so`, `libc++.so` and `libc++abi.so`, too. Please add `musl` `llvm-libs` to `depends`, too, if you want to clean everything up.

PKGBUILD
```diff
@@ -1,36 +1,40 @@
...
-sha256sums=('e36592e97fc4613a46b71f661e57c118aab20d451f92733728c00b5a45aa57a2'
-	'257fd00dc1f6ba7e79b70604aa358e481cff60b318de8d78b829acdec56f7eff')
+makedepends=(git meson ninja scdoc)
+source=(
+  git+$url.git#tag=v$pkgver
+  turnstiled.service
+)
+sha256sums=(
+  829cc9c01603e7de30b9307226d68da33f6e69d7db5fbf4c4dd578b0546ed482
+  34daaa1973722a835f4877f0e56598b6bea7d804cb44f9a2b6a9438d1820a02d
```
ziyao233: Please avoid reformatting sha256sums, it is usually automatically generated by `updpkgsums` (provided by pacman-contrib). Reformatting the array only brings unnecessary changes, which will be reverted back at the next time the package is updated through workflow.

PKGBUILD
```diff
@@ -1,36 +1,40 @@
...
+)
 
 build () {
   ewe-meson $pkgname build -D manage_rundir=true
-  meson compile -C build
+  ninja -C build
```
ziyao233: Why replace all calls to `meson` with `ninja`?

turnstiled.service
```diff
@@ -1,5 +1,5 @@
-type       = process
-before     = login.target
-depends-on = rc.target
-command    = /usr/bin/turnstiled
-logfile    = /var/log/turnstiled.log
+type = process
+command = /usr/bin/turnstiled
+logfile = /var/log/turnstiled.log
+before: login.target
+depends-on: rc.target
```
ziyao233: Please avoid re-spacing the unrelated lines. This brings unnecessary changes which is hard to blame, and doesn't even make the file easier to read.

ziyao233 (changes requested):

PKGBUILD
WeirdGumi: > Why replace all calls to `meson` with `ninja`?

I will revert it. Should we remove `ninja` from makedepends if we're using the `meson` calls?

ziyao233: meson hard depends on `ninja` as build tool, meson compile/test/install in turn invokes it.

So the answer is, yes, you could remove `ninja` from makedepends, but it doesn't mean ninja isn't necessary to build the package, it's just b.c. ninja has been listed as a dependency of meson, and it's redundant to specify it again here.

PKGBUILD
WeirdGumi: https://github.com/eweOS/packages/pull/6139/commits/cb40d5ab8334aa607f9af061bf44767a138d9f3a

PKGBUILD
WeirdGumi: https://github.com/eweOS/packages/pull/6139/commits/b91c069b998903bca8f502854f7e2a09f0ba2a05

PKGBUILD
WeirdGumi: https://github.com/eweOS/packages/pull/6139/commits/b91c069b998903bca8f502854f7e2a09f0ba2a05

PKGBUILD
WeirdGumi: https://github.com/eweOS/packages/pull/6139/commits/b91c069b998903bca8f502854f7e2a09f0ba2a05

turnstiled.service
WeirdGumi: https://github.com/eweOS/packages/pull/6139/commits/4211e0598a77f7936f0a9b2c8654486c229e075e

PR #6486, opened 2026-01-26
[wine] 11.1-1: new upstream version

YukariChiba:


binfmt.conf
```diff
@@ -0,0 +1,7 @@
+name: wine
+type: M
+offset: 0
+magic: MZ
+mask: 0
```
ziyao233: Should be

```
magic: 4d5a
mask: ffff
```

ziyao233 (changes requested):

ziyao233 (approved):

PR #6543, opened 2026-02-02
[appstream] 1.1.2-1: new upstream version

ziyao233:
Also enables --print-errorlogs for testing.


ziyao233:
Recheck after #6544 #6542

PR #6637, opened 2026-02-20
[webkit2gtk] 2.50.5-1: init package

ziyao233:


ziyao233:
Recheck after #6636, #6635, #6634, and #6633

PR #6807, opened 2026-03-11
[flashrom] 1.7.0-1: new upstream version, enable more features

ziyao233:
Enable libjaylink and libpci. Switch to Git sources.


ziyao233:
Recheck after #6806

PR #6991, opened 2026-04-02
[orc] 0.4.42-1: init package

YukariChiba:


PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+arch=(x86_64 aarch64 riscv64 loongarch64)
+license=(BSD-3-Clause)
+depends=(
+  musl
+)
+makedepends=(
+  git
+  linux-uapi-headers
+  meson
+  valgrind
```
ziyao233: I guess it's only for testing? We don't have `valgrind` on riscv64 and loongarch64, and this causes the package to fail to build on these architectures, so probably we should only include the dependency for x86_64 and aarch64.

PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+check() {
+  meson test -C build --print-errorlogs
+}
+
+package() {
+  provides=(liborc{,-test}-${pkgver%.*}.so)
+
+  meson install -C build --destdir "$pkgdir"
+  install -Dt "$pkgdir/usr/share/licenses/$pkgname" -m644 orc/COPYING
+}
```
ziyao233, line 38: Would `_install_license_` help here?

ziyao233 (changes requested):

PR #6997, opened 2026-04-03
[nodejs] 25.8.0-2: remove vendored ada-url and zstd

Integral-Tech:
Link against the system-provided ada-url and zstd instead of the bundled copies.


ziyao233 (approved):

ziyao233:
Thank you for your contribution!

PR #7002, opened 2026-04-05
[util-linux] 2.42-1: new upstream version

ziyao233:


ziyao233:
`STATX_MNT_ID` and its friends are added in Linux 6.11, and are introduced into musl since v1.2.6 (23ab04a86302 ("statx: add new struct statx fields and corresponding mask macros")
).

Recheck after musl v1.2.6 is ready.

PR #7102, opened 2026-04-12
[mesa] 26.0.4-2: add nvk dirver

panxi39:


PKGBUILD
```diff
@@ -89,7 +94,12 @@ build()
     -Ddefault_library=shared \
     -Dllvm-orcjit=true \
     -Dintel-rt=disabled \
-    -Dgallium-mediafoundation=disabled
+    -Dgallium-mediafoundation=disabled 
+
+  RUSTC_TARGET_LIBDIR="$(rustc --print target-libdir)"
+  echo "> Cleaning up self-contained paths from Ninja (Target: $RUSTC_TARGET_LIBDIR)..."
```
ziyao233: Please explain why this is necessary in comment.

PKGBUILD
```diff
@@ -89,7 +94,12 @@ build()
...
+  sed -i "s|-L$RUSTC_TARGET_LIBDIR/self-contained||g" build/build.ninja
```
ziyao233: Please don't use sed in build scripts, its behavior isn't predictable.

PKGBUILD
```diff
@@ -64,8 +67,10 @@ build()
 	    ;;
   esac
   VULKAN_LAYER=device-select,intel-nullhw,overlay,screenshot,vram-report-limit
-  ewe-meson $pkgbase-$pkgver build \
+
+  ewe-meson --wrap-mode nofallback --auto-features auto $pkgbase-$pkgver build \
```
ziyao233: AFAIK, the nvk backend requires downloading Rust crates at build-time, which would break offline builds. Please at least move downloading steps into `prepare()`, or ideally, make them inputs listed in `sources` so we could track them easily.

ziyao233: Why we need `--auto-features auto`? This overrides with the default (`--auto-features enabled`) set in `ewe-meson`.

If some features couldn't be enabled cuz of missing dependencies, etc., please explicitly disable them for clearance.

ziyao233: I'm not sure whether the meson scripts shipped by mesa correctly locks Rust crates versions/checksums, which is necessary for reproducible builds.

If it doesn't, the only way we could go is to manually track all checksums/versions of Rust crates in `sources` and `checksum`. Simply moving download steps to `prepare()` isn't an option in this case.

ziyao233 (changes requested):

PKGBUILD
```diff
@@ -28,23 +29,50 @@ makedepends=(
...
 # mold may fails with lto enabled
 options=(!lto)
 source=(
   https://mesa.freedesktop.org/archive/$pkgbase-$pkgver.tar.xz
+  "syn-2.0.87.tar.gz::https://crates.io/api/v1/crates/syn/2.0.87/download"
+  "quote-1.0.35.tar.gz::https://crates.io/api/v1/crates/quote/1.0.35/download"
+  "proc-macro2-1.0.86.tar.gz::https://crates.io/api/v1/crates/proc-macro2/1.0.86/download"
+  "unicode-ident-1.0.12.tar.gz::https://crates.io/api/v1/crates/unicode-ident/1.0.12/download"
+  "paste-1.0.14.tar.gz::https://crates.io/api/v1/crates/paste/1.0.14/download"
+  "rustc-hash-2.1.1.tar.gz::https://crates.io/api/v1/crates/rustc-hash/2.1.1/download"
```
ziyao233: It'd be better to have some variables like `_crate_syn_ver`/`_crate_quote_ver` tracking versions of the crates, so we don't need to touch two places in case that one crate needs updating.

PKGBUILD
```diff
@@ -28,23 +29,50 @@ makedepends=(
...
+  local subprojects_cache="$srcdir/$pkgbase-$pkgver/subprojects/packagecache"
+  mkdir -p "$subprojects_cache"
+
+  # Link downloaded Rust libraries to the subprojects cache
+  ln -sf "$srcdir/syn-2.0.87.tar.gz" "$subprojects_cache/"
+  ln -sf "$srcdir/quote-1.0.35.tar.gz" "$subprojects_cache/"
+  ln -sf "$srcdir/proc-macro2-1.0.86.tar.gz" "$subprojects_cache/"
+  ln -sf "$srcdir/unicode-ident-1.0.12.tar.gz" "$subprojects_cache/"
+  ln -sf "$srcdir/paste-1.0.14.tar.gz" "$subprojects_cache/"
+  ln -sf "$srcdir/rustc-hash-2.1.1.tar.gz" "$subprojects_cache/"
```
ziyao233:
```
ln -sf "$srcdir"/*.tar.gz "$subprojects_cache/"?
```

panxi39:
I have removed `sed` and `--wrap-mode nofallback --auto-features auto` in PKGBUILD, those problem has fixed after  meson patched (https://github.com/eweOS/packages/pull/7121).
And rust libs import has moved to `source()` and link it to subproject cache during `prepare()`.

PKGBUILD
```diff
@@ -28,23 +29,50 @@ makedepends=(
...
+}
```
ziyao233: Would setting `MESON_PACKAGE_CACHE_DIR`[1] to "$srcdir" when invoking meson be simpler?

[1]: https://github.com/mesonbuild/meson/blob/97802a56f302d1d97578b4b0b743f0c74fc72980/docs/markdown/Wrap-dependency-system-manual.md?plain=1#L117

ziyao233 (changes requested):

panxi39:
- Moved rust dependences to a list and use loop to add it to `source()`.
- Set Meson's subproject cache dir to `srcdir` to replace link file to original cache dir.

ziyao233 (approved):

ziyao233:
Thanks for your contribution!

PR #7143, opened 2026-04-17
[rust] 1.94.1-2: remove vendored libssh2 and libsqlite3

Integral-Tech:
Link against the system-provided libssh2 and libsqlite3 instead of the bundled copies.


ziyao233 (approved):

ziyao233:
Thanks for your contribution!

PR #7200, opened 2026-04-29
[cargo-c] 0.10.21-2: remove vendored libssh2 and libsqlite3, enable LTO

Integral-Tech:
- Link against the system-provided libssh2 and libsqlite3 instead of the bundled copies.
- Enable link-time optimization (LTO) and drop `-ffat-lto-objects` CFLAGS.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=cargo-c
 pkgver=0.10.21
-pkgrel=1
+pkgrel=2
 pkgdesc='A cargo subcommand to build and install C-ABI compatibile dynamic and static libraries'
 arch=(x86_64 aarch64 riscv64 loongarch64)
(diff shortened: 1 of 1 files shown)


ziyao233 (changes requested):
LTO works even without devendoring. eweOS by default utilizes Clang, which generates LLVM bitcode as well as Rustc, making cross-language LTO possible. Please fix the commit message.

Integral-Tech:
> LTO works even without devendoring. eweOS by default utilizes Clang, which generates LLVM bitcode as well as Rustc, making cross-language LTO possible. Please fix the commit message.

I've just pushed a new commit to fix the commit message.

ziyao233 (approved):

ziyao233:
Thanks for your contribution!

PR #7224, opened 2026-05-09
[nodejs] 25.8.0-3: switch to ninja build system

Integral-Tech:
Building Node.js using [Ninja](https://ninja-build.org) is faster than using [GNU Make](https://www.gnu.org/software/make/).

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -2,13 +2,13 @@
 
 pkgname=nodejs
 pkgver=25.8.0
-pkgrel=2
+pkgrel=3
 pkgdesc='JavaScript runtime'
 arch=(x86_64 aarch64 riscv64 loongarch64)
 url='https://nodejs.org/'
 license=('MIT')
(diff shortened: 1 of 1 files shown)


ziyao233:
Are there any benchmark results demonstrating the improvements?

Integral-Tech:
> Are there any benchmark results demonstrating the improvements?

Before: image


After:  image

ziyao233:
Tasks might be dispatched to workers with different performance levels, so I don't think this comparison makes much sense. I'll give it a try locally.

ziyao233:
On the same local machine, with the patch,

```
real    26m12.876s
user    390m34.788s
sys     41m50.966s
```

without the patch,

```
real    41m42.511s
user    392m13.776s
sys     40m45.969s
```

So it does improve the build time, while the user/sys part remains similar, thus I guess it's a concurrency issue.

Anyway, thanks for your contribution!

ziyao233 (approved):

PR #7233, opened 2026-05-18
[rust] 1.94.1-4: use system libgit2

YukariChiba:


ziyao233 (commented):
Considering forcing this with `LIBGIT2_NO_VENDOR=1` environment variable?

ziyao233 (dismissed):

PR #7757, opened 2026-07-05
[mpv] 0.41.0-5: Temporarily disable testsuite

ziyao233:
LLVM 21 has a new regression which causes OrcJIT fails to initialize with

	JIT session error: could not register eh-frame: __register_frame function not found

This breaks LLVMpipe, making it impossible to run testsuite on our builders, which has no access to hardware accelerated graphics API.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=mpv
 pkgver=0.41.0
-pkgrel=4
+pkgrel=5
 pkgdesc='a free, open source, and cross-platform media player'
 arch=(x86_64 aarch64 riscv64 loongarch64)
 license=('GPL-2.0-or-later')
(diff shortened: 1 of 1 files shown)


Icenowy:
Ah? Do this have a smallest reproducer?

ziyao233:
> Ah? Do this have a smallest reproducer?

This is an issue specific to compiler-rt systems, where `__register_frame` is provided by `libunwind.so` thus requires additional linking parameters, while LLVM's build system fails to account for this :(

We have a patch in LLVM[1] for solving similar issues back in LLVM 20 era. but it's getting rotten with LLVM 21.

[1]: https://github.com/llvm/llvm-project/pull/112087

Icenowy:
Well it looks like it's LLVM side issue instead of Mesa side.

PR #7763, opened 2026-07-05
[harbuzz] 14.2.1-1: new upstream version

ziyao233:


ziyao233:
~~Need testing on riscv64 and loongarch64.~~ Done and everything seems good.

PR #7936, opened 2026-07-18
[libpsl] 0.23.0-1: new upstream version

ziyao233:
Don't run autoreconf anymore since this project seems to have a (at least partially) hand-written configure script.


ziyao233:
Recheck after #7935

PR #7937, opened 2026-07-18
[curl] 8.21.0-2: Enable libpsl and remove unnecessary dependencies from ca-certs

ziyao233:


ziyao233:
I prefer to delay the merge until #7936 is merged.

PR #7941, opened 2026-07-21
[busybox] 1.37.0-16: change tun node to root:root 0666

WeirdGumi:


ziyao233 (approved):
Makes sense. Let's merge it after the CI finishes.

ziyao233:
Let's go.

PR #7958, opened 2026-07-26
[zig] 0.16.0-1: init package

WeirdGumi:


PKGBUILD
```diff
@@ -0,0 +1,40 @@
+# Maintainer: Weird Gumi <weirdgumi@tutamail.com>
+
+pkgname=zig
+pkgver=0.16.0
+pkgrel=1
+pkgdesc='A general-purpose programming language and toolchain for maintaining robust, optimal and reusable software'
```
ziyao233: https://os-wiki.ewe.moe/en/dev/topic/packaging/pkgdesc-guideline

> pkgdesc shouldn't be an A.D. for the project.

Consider something like `Compiler and runtime for the Zig language` or `Toolchain for the Zig language`, I'm not sure which is better.

PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url=https://ziglang.org
+license=(MIT)
+depends=(clang llvm llvm-libs musl zlib zstd)
```
ziyao233: Please use `zlib-ng` instead. Why is clang dependent on? If it requires a C compiler to work, is an external linker is also necessary? If so please include mold in `depends`, too.

ziyao233: Another question, which LLVM versions is Zig compatible with? We have soon a plan to upgrade to LLVM 22, and so far there's no mechanism for multiple LLVM versions to co-exist.

PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+    ZIG_VERSION=$pkgver
+    ZIG_PIE=ON
+    ZIG_RELEASE_SAFE=ON
+  )
+  cmake -S $pkgname-$pkgver -B build ${cmake_vars[@]/#/-D} -G Ninja
+  cmake --build build
+}
+
+package() {
+  DESTDIR=$pkgdir cmake --install build
```
ziyao233: Please quote `$pkgdir`.

PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+
+pkgname=zig
+pkgver=0.16.0
+pkgrel=1
+pkgdesc='A general-purpose programming language and toolchain for maintaining robust, optimal and reusable software'
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url=https://ziglang.org
+license=(MIT)
+depends=(clang llvm llvm-libs musl zlib zstd)
+makedepends=(lld llvm-devel ninja)
```
ziyao233: Doesn't it depend on CMake at build-time, too?

ziyao233 (changes requested):

PKGBUILD
WeirdGumi: I'll fix this. I thought our base-devel included cmake.

PKGBUILD
WeirdGumi: Zig 0.16.0 only supported LLVM 21. but the master branch uses LLVM 22, so it might be supported in the next version.

WeirdGumi: https://github.com/ziglang/zig/issues/24542

WeirdGumi: Zig needs libclang-cpp.so to work. Regarding the linker part, Zig itself is a linker and also embeds LLD, which can be switched.

ziyao233 (approved):

PR #7966, opened 2026-07-27
[f2fs-tools] 1.16.0-1: init package

WeirdGumi:


PKGBUILD
```diff
@@ -0,0 +1,45 @@
...
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url=https://git.kernel.org/pub/scm/linux/kernel/git/jaegeuk/$pkgname.git/about
+license=('GPL-2.0-only AND (GPL-2.0-only OR LGPL-2.1-only)')
+depends=(lz4 musl util-linux-libs)
+makedepends=(bsd-compat-headers linux-uapi-headers)
+provides=(libf2fs.so libf2fs_format.so)
+source=(
+  https://git.kernel.org/pub/scm/linux/kernel/git/jaegeuk/$pkgname.git/snapshot/$pkgname-$pkgver.tar.gz
+  convert-to-lseek-and-kill-lseek64.patch
+  link-using-libf2fs.la.patch
```
ziyao233: Are these patches taken from somewhere? Please add a note about the origin if yes.

PKGBUILD
```diff
@@ -0,0 +1,45 @@
...
+
+check() {
+  cd $pkgname-$pkgver
+  make check
+}
+
+package() {
+  cd $pkgname-$pkgver
+  make install DESTDIR="$pkgdir" SBINDIR=/usr/bin
+  _install_license_ COPYING
```
ziyao233: GPL licenses don't require a copy of the license to be shipped together, but this isn't a requirement, and it totally depends on you whether to do so.

convert-to-lseek-and-kill-lseek64.patch
ziyao233, line 1: This patch is quite large and complex, and is probably hard to maintain. Could you explain what it's trying to fix, so we might come up with a better way? It deserves a comment in the PKGBUILD, too.

convert-to-lseek-and-kill-lseek64.patch
```diff
@@ -0,0 +1,458 @@
+From b15b6cc56ac7764be17acbdbf96448f388992adc Mon Sep 17 00:00:00 2001
+From: Waldemar Brodkorb <wbx@openadk.org>
+Date: Wed, 23 Aug 2023 15:41:28 +0200
+Subject: f2fs-tools: convert to lseek() and kill lseek64
+
+This patch replaces lseek64 with lseek() having #define _FILE_OFFSET_BITS 64.
```
ziyao233: However it also removes some detection against llseek in configuration.ac, please explain why, and why it's okay.

ziyao233: A simpler way to deal with 64-bit file APIs is to define macro `_LARGEFILE64_SOURCE`. This instructs musl to provide compatibility macros for it, and would probably make the patch easier to maintain.

ziyao233 (changes requested):

ziyao233:
Generally this PR looks good, but I want to simplify the first patch to ease future maintenance.

PKGBUILD
WeirdGumi: Backport from f2fs-tools repository

WeirdGumi: https://git.kernel.org/pub/scm/linux/kernel/git/jaegeuk/f2fs-tools.git/commit/?id=b15b6cc56ac7764be17acbdbf96448f388992adc

https://git.kernel.org/pub/scm/linux/kernel/git/jaegeuk/f2fs-tools.git/commit/?id=ab4df39dcddf84429060d47f63570ca5704104d4

convert-to-lseek-and-kill-lseek64.patch
WeirdGumi: These patches are from upstream and do not require maintenance.

PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+makedepends=(bsd-compat-headers linux-uapi-headers)
+provides=(libf2fs.so libf2fs_format.so)
+source=(
+  https://git.kernel.org/pub/scm/linux/kernel/git/jaegeuk/f2fs-tools.git/snapshot/$pkgname-$pkgver.tar.gz
+
+  # These patches are backports from upstream.
+  # https://git.kernel.org/pub/scm/linux/kernel/git/jaegeuk/f2fs-tools.git/commit/?id=b15b6cc56ac7764be17acbdbf96448f388992adc
+  convert-to-lseek-and-kill-lseek64.patch
+
+  # https://git.kernel.org/pub/scm/linux/kernel/git/jaegeuk/f2fs-tools.git/commit/?id=ab4df39dcddf84429060d47f63570ca5704104d4
```
ziyao233: Comments inside an array may break OBS' PKGBUILD parser. Please avoid this.

ziyao233 (changes requested):

ziyao233 (approved):

PR #7970, opened 2026-07-27
[zsh] 5.9.2-1: new upstream version

WeirdGumi:


PKGBUILD
```diff
@@ -1,66 +1,66 @@
...
-makedepends=('linux-headers')
-source=("https://www.zsh.org/pub/zsh-${pkgver}.tar.xz"
-  zsh-59-clang-15-configure.patch)
-sha512sums=('d9138b7f379ad942a5f46819d2dd52d31f3a1129f2a0d1b53d4c5cd43c318b60396da6d37c57c477b8e958fb750209aca0ae93f8c9dd42ac958de006a0ff067e'
-            '130b9b829200b7ad49b87dc1d3a549ca3a36aba57f01f9d4ef60ba66d8f392f57f4d6a3d35ca323857b27d343c9e6218ffeca233fc436b74c91cf88c915146d8')
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url=https://www.zsh.org
+license=('LicenseRef-Zsh AND GPL-2.0-only')
+depends=(gdbm libcap musl ncurses)
+makedepends=(linux-headers)
```
ziyao233: This name is provided for compatibility purpose, please use linux-uapi-headers instead.

PKGBUILD
```diff
@@ -1,66 +1,66 @@
...
 
-  # patch configure
-  patch -p1 < $srcdir/zsh-59-clang-15-configure.patch
-  autoreconf -fiv
+  cd $pkgname-$pkgver
+  rm -r Completion/{AIX,BSD,Cygwin,Darwin,Debian,Mandriva,openSUSE,Redhat,Solaris,X}
 
-  # remove failed tests
-  rm -rf Test/{A01grammar,V09datetime}.ztst
+  # There tests fail without --disable-multibyte, so remove them.
```
ziyao233: Could you investigate why? It's not a hard requirement for this to merge.

PKGBUILD
```diff
@@ -1,66 +1,66 @@
...
+    --enable-zshrc=/etc/$pkgname/zshrc \
+    --enable-zprofile=/etc/$pkgname/zprofile \
+    --enable-zlogin=/etc/$pkgname/zlogin \
+    --enable-zlogout=/etc/$pkgname/zlogout \
+    --enable-fndir=/usr/share/$pkgname/functions \
     --enable-function-subdirs \
-    --enable-fndir=/usr/share/zsh/functions \
-    --enable-scriptdir=/usr/share/zsh/scripts \
-    --with-tcsetpgrp \
+    --enable-scriptdir=/usr/share/$pkgname/scripts \
```
ziyao233: I don't think it helps much to replace the hard-coded `zsh` with `$pkgname`, but only makes the command line harder to read. Please keep the current form, and it would be also easier to blame the changelog.

0002-fix-keymaps-path.patch
```diff
@@ -0,0 +1,11 @@
+--- a/Completion/Unix/Command/_loadkeys
++++ b/Completion/Unix/Command/_loadkeys
+@@ -9,7 +9,7 @@
+       '(--mktable -m)'{--mktable,-m}'[output a "defkeymap.c" to stdout]' \
+       '(--clearstrings)'{--clearstrings,-s}'[clear kernel string table]' \
+       '*'{--verbose,-v}'[print more details]' \
+-      ':keymap:_files -W /usr/share/keymaps'
++      ':keymap:_files -W /usr/share/kbd/keymaps'
```
ziyao233, line 8: Could you explain what this patch is for?

PKGBUILD
```diff
@@ -1,66 +1,66 @@
...
+license=('LicenseRef-Zsh AND GPL-2.0-only')
+depends=(gdbm libcap musl ncurses)
+makedepends=(linux-headers)
+backup=(etc/$pkgname/zprofile)
+install=$pkgname.install
+source=(
+  $url/pub/$pkgname-$pkgver.tar.xz
+  zprofile
+  drop-completions.patch
+  fix-keymaps-path.patch
```
ziyao233: Please add comments to inform the origin of these patches and purposes. It could "Downstream" if it's written by you.

PKGBUILD
```diff
@@ -1,66 +1,66 @@
...
+    --enable-fndir=/usr/share/$pkgname/functions \
     --enable-function-subdirs \
-    --enable-fndir=/usr/share/zsh/functions \
-    --enable-scriptdir=/usr/share/zsh/scripts \
-    --with-tcsetpgrp \
+    --enable-scriptdir=/usr/share/$pkgname/scripts \
     --enable-pcre \
     --enable-cap \
-    --enable-zsh-secure-free
+    --enable-gdbm
```
ziyao233, line 56: Just to check, all these removed options are removed in this new version of zsh, aren't they?

ziyao233 (changes requested):

0002-fix-keymaps-path.patch
WeirdGumi: the patch corrects the completion path for `loadkey` in the `kbd` package to the proper location.

PKGBUILD
WeirdGumi: `--with-tcsetpgrp` is removed, and other...
`--enable-multibyte` does not need to be explicitly set.
`--enable-zsh-secure-free` don't know what is for, it seems to be debug related.
`--with-term-lib` may not need to be explicitly set.

PKGBUILD
```diff
@@ -1,66 +1,69 @@
...
+source=(
+  $url/pub/$pkgname-$pkgver.tar.xz
+  zprofile
 
-prepare()
-{
-  cd $pkgname-$pkgver
-  sed -i 's#/usr/share/keymaps#/usr/share/kbd/keymaps#g' Completion/Unix/Command/_loadkeys
-  sed -i 's#/usr/share/misc/usb.ids#/usr/share/hwdata/usb.ids#g' Completion/Linux/Command/_lsusb
+  # These patches are from downstream.
```
ziyao233: Comments inside an array breaks OBS' parser. A good idea is to number the patches (`0001-....patch`) and add a block of comments above the sources array explaining their origins and purposes.

ziyao233 (changes requested):

ziyao233 (approved):

PR #7972, opened 2026-07-27
[yq] 4.53.3-1: init package

WeirdGumi:


PKGBUILD
```diff
@@ -0,0 +1,38 @@
+# Maintainer: Weird Gumi <weirdgumi@tutamail.com>
+
+pkgname=yq
+pkgver=4.53.3
+pkgrel=1
+pkgdesc='A portable command-line YAML, JSON, XML, CSV, TOML, HCL and properties processor'
+arch=(x86_64 aarch64 riscv64 loongarch64)
+url=https://mikefarah.gitbook.io/yq
+license=(MIT)
+depends=(musl)
```
ziyao233, line 10: Usually Go programs don't depend on musl, unless CGO is used. Is it the case? Please check again with `readelf --needed-libs`.

PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+}
+
+package() {
+  cd $pkgname-$pkgver
+  install -Dt "$pkgdir"/usr/bin $pkgname
+  install -Dm644 -t "$pkgdir"/usr/share/doc/$pkgname README.md
+  _install_license_ LICENSE
+  ./$pkgname shell-completion bash | install -Dm644 /dev/stdin "$pkgdir"/usr/share/bash-completion/completions/$pkgname
+  ./$pkgname shell-completion zsh | install -Dm644 /dev/stdin "$pkgdir"/usr/share/zsh/site-functions/_$pkgname
+  ./$pkgname shell-completion fish | install -Dm644 /dev/stdin "$pkgdir"/usr/share/fish/vendor_completions.d/$pkgname.fish
```
ziyao233: I think these commands might be clear if you use `yq` directly instead of referencing `$pkgname`. This isn't a hard requirement, it depends on you whether to change.

ziyao233 (changes requested):

PKGBUILD
WeirdGumi:
```
NeededLibraries [
  libc.so
]
```

PKGBUILD
WeirdGumi: done.

ziyao233 (approved):

PR #7977, opened 2026-07-29
[meson] 1.11.2-2: correct license, switch to python-build, add missing dep, adjust ewe-meson flags

WeirdGumi:


ewe-meson
```diff
@@ -5,8 +5,8 @@ exec meson setup \
   --prefix        /usr \
   --libexecdir    lib \
   --sbindir       bin \
-  --includedir    include \
   --buildtype     plain \
-  --auto-features enabled \
```
ziyao233, line 10: Why?

ewe-meson
```diff
@@ -5,8 +5,8 @@ exec meson setup \
   --prefix        /usr \
   --libexecdir    lib \
   --sbindir       bin \
-  --includedir    include \
```
ziyao233, line 8: So this is the default, right?

ziyao233 (changes requested):

ewe-meson
WeirdGumi: Yes. You can check `meson setup --help`

ewe-meson
WeirdGumi: removed `--auto-features enabled` to let `meson` fall back to its default auto-detection. this avoids having to explicitly set features that are unavailable or missing dependencies as disabled.

ziyao233: Ok, so it's desired to set --auto-features to enabled. We do want to enable as much features as possible and only opt-out them by default, please don't change the default.

ziyao233 (approved):

PR #7986, opened 2026-07-29
[libffi] 3.7.1-1: new upstream version

ziyao233:


ziyao233:
This has been tested on riscv64 with python-gobject's testsuite, and no regression has been observed. Removing the patch is okay.

PR #8003, opened 2026-08-05
[river] 0.4.8-1: init package

WeirdGumi:


PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+  cd $pkgname-$pkgver
+  DESTDIR=build zig build \
+    -Dtarget=$CARCH-linux-musl \
+    -Dcpu=baseline \
+    -Doptimize=ReleaseSafe \
+    -Dstrip \
+    -Dpie \
+    -Dllvm \
+    -Dman-pages \
+    --search-prefix /usr
```
ziyao233: `check()` duplicates these arguments, please unify them into a file-level array so they're easier to track and modify.

PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+}
+
+check() {
+  cd $pkgname-$pkgver
+  zig build test \
```
ziyao233: What's the purpose of `DESTDIR`, and why it's only set in `build()`?

ziyao233 (changes requested):

PKGBUILD
WeirdGumi: `zig build` implicitly runs `install`. Without `DESTDIR`, it outputs to `./zig-out` and embeds that absolute path as the prefix. With `DESTDIR=build`, the prefix becomes `/usr` and files land in `build/usr` — exactly what `package()` expects.

`check()` runs `zig build test`, which only tests, so `DESTDIR` is unnecessary there.

PKGBUILD

ziyao233 (approved):

PR #8006, opened 2026-08-07
[wmenu] 0.2.0-3: update sha256sum, fix depends, add missing license file, quote variable $pkgdir

WeirdGumi:


PKGBUILD
```diff
@@ -2,15 +2,15 @@
...
 url='https://codeberg.org/adnano/wmenu'
 arch=(x86_64 aarch64 riscv64 loongarch64)
 license=(MIT)
-depends=(cairo pango wayland wayland-protocols libxkbcommon)
-makedepends=(meson samurai scdoc)
+depends=(cairo glib libxkbcommon musl pango wayland)
+makedepends=(meson scdoc wayland-protocols)
 source=("$url/archive/$pkgver.tar.gz")
-sha256sums=('4e6aea3f8975fec720f6eb87aad620d5297a8a5a137615e4cf047e95d2b9d308')
+sha256sums=('b1b593a80310085a90bd349d458ffbcfacf875d74468ae4428bdf5692a671ef9')
```
ziyao233, line 13: So the checksum differs while the `pkgver` isn't updated? It might be an indication that codeberg doesn't guarantee to generate reproducible tarballs, please consider building from manually uploaded artifacts, or from git instead.

ziyao233 (changes requested):

PKGBUILD
WeirdGumi: Check: https://codeberg.org/Codeberg/Community/issues/2861

ziyao233: Please mention this in the commit message, providing the link, thanks.

WeirdGumi: Done.

ziyao233 (approved):

PR #8007, opened 2026-08-07
[leveldb] 1.23-1: init package

WeirdGumi:


PKGBUILD
```diff
@@ -0,0 +1,54 @@
...
+  cmake -S $pkgbase-$pkgver -B build ${cmake_vars[@]/#/-D} -G Ninja
+  cmake --build build
+}
+
+check() {
+  ctest --output-on-failure --test-dir build
+}
+
+package_leveldb() {
+  optdepends=('leveldb-doc: Documentation')
```
ziyao233: I don't think it's appropriate to list it as an optional dependency, since it doesn't improve or extend leveldb's functionality.

ziyao233 (changes requested):

PKGBUILD

ziyao233 (approved):

PR #8010, opened 2026-08-07
[rime-essay] 20260713.e9b1a374-1: init package

WeirdGumi:


PKGBUILD
```diff
@@ -0,0 +1,16 @@
+# Maintainer: Weird Gumi <weirdgumi@tutamail.com>
+
+pkgname=rime-essay
+pkgver=20260713
```
ziyao233: Please include the commit hash abbreviated to 8 or 12 characters in pkgver, so it's more informative. e.g. `20260509.082425ea`. You could truncate the `$_commit` variable with bash variable substitution.

ziyao233 (changes requested):

PKGBUILD

ziyao233 (approved):

PR #8013, opened 2026-08-09
[python-packaging] 26.3-1: new upstream version

WeirdGumi:


ziyao233 (commented):
Please read and check the changelog before upgrading a Python package: does the new version bring in compatibility issues? If so, please test against some existing packages to see if we're affected.

PR #8063, opened 2026-08-13
[c3c] 0.8.3-1: init package

WeirdGumi:


0001-add-arm-disable.patch
```diff
@@ -0,0 +1,23 @@
...
+@@ -573,6 +574,12 @@
+         ${CMAKE_BINARY_DIR}/git_hash.h
+         ${CMAKE_BINARY_DIR}/docs_template.h
+ )
++
++# Building C3 with llvm toolchains that don't have ARM Support
++if(C3_ARM_DISABLE)
++    message(STATUS "ARM support is disabled")
++	target_compile_definitions(c3c PRIVATE ARM_DISABLE)
++endif()
```
ziyao233, line 20: So this is because eweOS doesn't enable ARM support for LLVM, right? If please comment about the reason.

ziyao233 (changes requested):

0001-add-arm-disable.patch
WeirdGumi: Yup.
https://github.com/eweOS/packages/blob/llvm/PKGBUILD#L237

ziyao233 (approved):

PR #8068, opened 2026-08-13
[font-tinos] 1.340-1: init package

WeirdGumi:


ziyao233:
I don't think the commit states the version is 1.340 in any way? If so, please treat it as a Git-hash-based package, and include the commit hash in the pkgver.

WeirdGumi:
> I don't think the commit states the version is 1.340 in any way? If so, please treat it as a Git-hash-based package, and include the commit hash in the pkgver.

https://github.com/googlefonts/tinos/commit/6a938e5f172bb281af077ee11b452b97e03395d8

ziyao233:
> > I don't think the commit states the version is 1.340 in any way? If so, please treat it as a Git-hash-based package, and include the commit hash in the pkgver.
> 
> [googlefonts/tinos@6a938e5](https://github.com/googlefonts/tinos/commit/6a938e5f172bb281af077ee11b452b97e03395d8)

However, this doesn't match the commit you specified in the PKGBUILD?

JulianDroske:
> > > I don't think the commit states the version is 1.340 in any way? If so, please treat it as a Git-hash-based package, and include the commit hash in the pkgver.
> > 
> > 
> > [googlefonts/tinos@6a938e5](https://github.com/googlefonts/tinos/commit/6a938e5f172bb281af077ee11b452b97e03395d8)
> 
> However, this doesn't match the commit you specified in the PKGBUILD?

---

Well, this is because of the poor management of this repo.

This repo has no releases, and the commit (also latest, `3b4482a`) follows up the previous "release" one (`6a938e5`) in the same day. It updated font binaries without touching the source code, both in one repo.

From my point of view, it is better to mark the real commit in either commit message or version code like `1.340.3b4482a`, but I would say this current version still make sense.

And I do not suggest it is so worth fetching the "release" commit and build manually to just keep the consistency?

ziyao233:
> > > > I don't think the commit states the version is 1.340 in any way? If so, please treat it as a Git-hash-based package, and include the commit hash in the pkgver.
> > > 
> > > 
> > > [googlefonts/tinos@6a938e5](https://github.com/googlefonts/tinos/commit/6a938e5f172bb281af077ee11b452b97e03395d8)
> > 
> > 
> > However, this doesn't match the commit you specified in the PKGBUILD?
> 
> Well, this is because of the poor management of this repo.
> 
> This repo has no releases, and the commit (also latest, `3b4482a`) follows up the previous "release" one (`6a938e5`) in the same day. It updated font binaries without touching the source code, both in one repo.
> 
> From my point of view, it is better to mark the real commit in either commit message or version code like `1.340.3b4482a`, but I would say this current version still make sense.

Ahh, thanks for the explanation. This now makes sense to me.

> And I do not suggest it is so worth fetching the "release" commit and build manually to just keep the consistency?

ziyao233 (approved):

PR #8176, opened 2026-08-27
[libglvnd] 1.7.0-7: use local meson options instead of MESONFLAGS, enable gles1, add missing dependency

WeirdGumi:


PKGBUILD
ziyao233: However, you forgot to update this reference to MESONFLAGS.

ziyao233 (changes requested):

PKGBUILD

ziyao233 (approved):

PR #8185, opened 2026-08-30
[mesa] 26.2.0-2: simplify meson options and packaging

WeirdGumi:


PKGBUILD
```diff
@@ -67,57 +67,52 @@ sha512sums=('c7c810c6958fa18eb75eea9968d84d0edd29b579d351a7f22d8c2b2b13b6e04e5b0
...
+    android-libbacktrace=disabled
+    valgrind=disabled
+    libunwind=disabled
+    lmsensors=disabled
+    default_library=shared
+    llvm-orcjit=true
+    intel-rt=disabled
+    gallium-mediafoundation=disabled
+  )
+
```
ziyao233, line 86: I don't think this large change improves maintainability or readability much, but it breaks bisectability. So I would NAK this.

ziyao233 (changes requested):

PKGBUILD
WeirdGumi: Okay.

PR #8226, opened 2026-09-03
[opencc] 1.4.2-1: new upstream version, package python bindings separately

WeirdGumi:


PKGBUILD
```diff
@@ -1,40 +1,62 @@
...
 arch=(x86_64 aarch64 riscv64 loongarch64)
 license=(Apache-2.0)
-depends=(libmarisa)
-makedepends=(cmake tclap python rapidjson)
+depends=(libmarisa llvm-libs musl)
+makedepends=(cmake tclap python-build python-installer python-setuptools rapidjson)
 checkdepends=(googletest)
-source=("https://github.com/BYVoid/OpenCC/archive/refs/tags/ver.$pkgver.tar.gz")
-sha256sums=('f4f86eb25e239450d075081e08594801aa063c298d21d9f6c6aa85cd55241962')
+# 0001: From downstream.
```
ziyao233: And it's purpose?

ziyao233 (changes requested):

PKGBUILD
```diff
@@ -1,40 +1,62 @@
...
 
-package() {
-	DESTDIR=$pkgdir cmake --install OpenCC-ver.$pkgver/build
+package_opencc() {
+	DESTDIR="$pkgdir" cmake --install build
 	_install_license_ OpenCC-ver.$pkgver/LICENSE
 }
+
+package_python-opencc() {
+	depends+=(python)
```
ziyao233, line 57: Doesn't it depend on `opencc` as well?

PKGBUILD
WeirdGumi: This patch makes the Python package reuse the files already built during the CMake stage instead of recompiling them.

ziyao233 (changes requested):

PKGBUILD
ziyao233: Please comment in the PKGBUILD, thanks.

PKGBUILD
WeirdGumi: It has its own opencv_clib.python-313-x86_64-linux-musl.so bundled.

PKGBUILD

ziyao233 (approved):

PR #8240, opened 2026-09-11
[c3fmt] 0.3.2-1: init package

WeirdGumi:


PKGBUILD
```diff
@@ -0,0 +1,24 @@
+# Maintainer: Weird Gumi <weirdgumi@tutamail.com>
+
+pkgname=c3fmt
+pkgver=0.3.2
+pkgrel=1
+pkgdesc='A customizable code formatter for the C3 language'
+arch=(x86_64 aarch64 riscv64)
```
ziyao233, line 7: Why is loongarch64 excluded?

ziyao233 (changes requested):

PKGBUILD
WeirdGumi: c3 doesn't support it right now.

YukariChiba (approved):

PR #8258, opened 2026-09-19
[pipewire] 1.6.9-2: Enable bluez5 integration

ziyao233:


ziyao233:
Recheck after #8259 #8260 #8261

